PR #1, opened 2018-12-21
Main classes

Flix6x:


timely_beliefs/tests/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+    psql -U postgres --password -h 127.0.0.1 -d tbtest
+    \q
+
+Create tables
+===================
+
+Run the following:
+
+    python timely_beliefs/tests/create_tables.py
```
nhoening: Why not let each test run create_all() and then drop_all(), like we did in bvp? We can also make a ticket for this.

Flix6x: Excellent suggestion. Done in commit ecd9b6e.

PR #2, opened 2019-01-22
Dataframe methods

Flix6x:


timely_beliefs/sensors/utils.py
```diff
@@ -0,0 +1,74 @@
...
+                try:
+                    d2[k] = parse_datetime(v)
+                except ISO8601Error:
+                    d2[k] = v
+        else:
+            d2[k] = v
+    return d2
+
+
+def func_store_list() -> dict:
```
nhoening: This is probably a general utility, not just one for sensors?

nhoening: Ah, I see inside that it only looks for knowledge horizon functions...

timely_beliefs/sensors/utils.py
```diff
@@ -0,0 +1,74 @@
...
+    """Returns a dictionary with function names and Callable objects supported in our function store."""
+    functions_dict = {o[0]: o[1] for o in getmembers(knowledge_horizons) if isfunction(o[1])}
+    return functions_dict
+
+
+def eval_verified_knowledge_horizon_fnc(fnc: Callable, par: dict, event_start: Optional[datetime]):
```
nhoening: fnc is a string, I believe, not a Callable. That confused me when reading further. You could also name `k` and `v` better, e.g. `fnc_name` and `fnc`

PR #4, opened 2019-03-15
feature/data import

nhoening:
Make it possible to import data (in a notebook, for instance). I split classes into normal ones and DB ones.


timely_beliefs/beliefs/utils.py
```diff
@@ -37,11 +56,25 @@ def replace_multi_index_level(df: "classes.BeliefsDataFrame", level: str, index:
...
     else:
         df[level] = index.to_series(name="").reset_index()[index.name]
     df = df.rename({level: index.name}, axis="columns")
 
     # Convert columns to index levels (only columns that represent index levels)
     return df.set_index(indices)
+
+
+def load_time_series(
+    series: pd.Series, sensor: Sensor, source: BeliefSource, horizon: timedelta
```
Flix6x: I suggest renaming `horizon` to `belief_horizon` and possibly also `series` to `event_value_series`. Would also be nice to be able to specify `belief_time` rather than `belief_horizon`.

timely_beliefs/beliefs/classes.py
```diff
@@ -175,37 +197,102 @@ class BeliefsDataFrame(DataFrame):
...
+        super().__init__(*args, **kwargs)
+
+        # Set the Sensor metadata (including timing properties of the sensor)
+        self.sensor = sensor
+
+    def add_from_time_series(self,
+        series: pd.Series, source: BeliefSource, horizon: timedelta
+    ):
+        """Add beliefs from time series entries into this TimedBeliefsDataFrame.
+        Append makes no sense with this class atm."""
```
Flix6x: Maybe `append()` does work now with the new code on master?

timely_beliefs/beliefs/classes.py
```diff
@@ -1,46 +1,49 @@
...
     """"""
 
-    @declared_attr
-    def __tablename__(cls):
-        return cls.__name__.lower()
+    event_start: datetime
+    belief_horizon: timedelta
+    event_value: float
+    sensor: Sensor
+    source: str  # TODO: needs to be an object as well?
```
Flix6x: You do initialize with the object, but convert to a string? If you really need to convert, why not store the source_id (i.e. an integer) similar to what we do in the BeliefsDataFrame? The init() function of the BeliefsDataFrame will look for the source_id attribute in any case.

timely_beliefs/beliefs/classes.py
nhoening: It does work, but it also cannot be an inplace operation. I still made it possible via the BeliefsDataFrame class.

PR #5, opened 2019-04-06
Accuracy calculations and fixed horizon beliefs (for flashlight plot)

Flix6x:
Already rebased on master.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -370,9 +370,15 @@ class BeliefsDataFrame(pd.DataFrame):
         )
 
     @hybrid_method
-    def belief_history(self, event_start) -> "BeliefsDataFrame":
+    def belief_history(self, event_start: datetime) -> "BeliefsDataFrame":
         """Select all beliefs about a single event, identified by the event's start time."""
-        return self.xs(event_start, level="event_start").sort_index()
+        return self.xs(enforce_utc(event_start), level="event_start").sort_index()
+
+    @hybrid_method
+    def fixed_horizon(self, belief_time: datetime) -> "BeliefsDataFrame" :
+        """Select the most recent belief about each event at a given belief time."""
+        df = self[self.index.get_level_values("belief_time") <= enforce_utc(belief_time)]
+        return belief_utils.select_most_recent_belief(df)
 
     @hybrid_method
     def rolling_horizon(self, belief_horizon: timedelta) -> "BeliefsDataFrame":
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -640,8 +640,15 @@ class BeliefsDataFrame(pd.DataFrame):
     ) -> "BeliefsDataFrame":
         """Simply get the accuracy of beliefs about events, at a given time (pass a datetime) or at a given horizon
         (pass a timedelta).
-        Optionally, set a reference to get the accuracy of beliefs with respect to those held by a specific source.
+
+        By default the accuracy is determined with respect to the most recent beliefs held by the same source.
+        Optionally, set a reference source to determine accuracy with respect to beliefs held by a specific source.
+
         By default the accuracy is determined with respect to the most recent beliefs.
+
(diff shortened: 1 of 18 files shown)


timely_beliefs/utils.py
```diff
@@ -10,3 +13,113 @@ def enforce_utc(dt: datetime) -> pd.Timestamp:
...
+            np.seterr(**old_settings)
+            return out
+
+        return wrapper
+
+    return decorator
+
+
+@with_error_settings(divide="ignore")
+def divide_ignore(*args, **kwargs):
```
nhoening: not in use, and then the above decroator might also not be needed?

timely_beliefs/utils.py
```diff
@@ -10,3 +13,113 @@ def enforce_utc(dt: datetime) -> pd.Timestamp:
...
+    return np.divide(*args, **kwargs)
+
+
+def replace_multi_index_level(
```
nhoening: Something's off because it seems to me this function exists twice:

```
± grep -r "def replace_multi_index_level" timely_beliefs
timely_beliefs/beliefs/utils.py:def replace_multi_index_level(
timely_beliefs/utils.py:def replace_multi_index_level(
```

timely_beliefs/beliefs/classes.py
```diff
@@ -351,36 +379,219 @@ def knowledge_horizons(self) -> pd.TimedeltaIndex:
...
+                df = df[
+                    df.index.get_level_values("belief_horizon")
+                    <= belief_horizon_window[1]
+                ]
+            df = df.convert_index_from_belief_horizon_to_time()
+        df.index = df.index.droplevel("event_start")
+        return df
+
+    @hybrid_method
+    def fixed_horizon(
```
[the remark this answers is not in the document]
Flix6x: I always have a mental itch when talking about the concepts of "fixed horizon" and "rolling horizon". Because the rolling horizon implies a constant horizon for each different event and the fixed horizon implies a changing horizon for each different event.

I agree that a name change is in order, but I think we can do better than changing `fixed_horizon` to `fixed_beliefs`, since the act of fixing something should pertain to timing more explicitly. Perhaps ` fixed_viewpoint`? And likewise, ` rolling_viewpoint` for `rolling_horizon`. So we'd get:

`fixed_viewpoint` -> the horizon increases as you look further ahead.
`rolling_viewpoint` -> the horizon stays the same, because your viewpoint moves with you as you look further ahead.

nhoening:
I like it

Am 19. April 2019 12:08:04 MESZ schrieb Felix Claessen <notifications@github.com>:
>Flix6x commented on this pull request.
>
>
>
>> +            if belief_horizon_window[0] is not None:
>+                df = df[
>+                    df.index.get_level_values("belief_horizon")
>+                    >= belief_horizon_window[0]
>+                ]
>+            if belief_horizon_window[1] is not None:
>+                df = df[
>+                    df.index.get_level_values("belief_horizon")
>+                    <= belief_horizon_window[1]
>+                ]
>+            df = df.convert_index_from_belief_horizon_to_time()
>+        df.index = df.index.droplevel("event_start")
>+        return df
>+
>+    @hybrid_method
>+    def fixed_horizon(
>
>I always have a mental itch when talking about the concepts of "fixed
>horizon" and "rolling horizon". Because the rolling horizon implies a
>constant horizon for each different event and the fixed horizon implies
>a changing horizon for each different event.
>
>I agree that a name change is in order, but I think we can do better
>than changing `fixed_horizon` to `fixed_beliefs`, since the act of
>fixing something should pertain to timing more explicitly. Perhaps `
>fixed_viewpoint`? And likewise, ` rolling_viewpoint` for
>`rolling_horizon`. So we'd get:
>
>`fixed_viewpoint` -> the horizon increases as you look further ahead.
>`rolling_viewpoint` -> the horizon stays the same, because your
>viewpoint moves with you as you look further ahead.
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/SeitaBV/timely-beliefs/pull/5#discussion_r276954114

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

PR #6, opened 2019-05-17
First visuals

Flix6x:


timely_beliefs/beliefs/classes.py
```diff
@@ -788,26 +857,102 @@ def rolling_viewpoint_accuracy(
...
-        >>> df.rolling_viewpoint_accuracy(belief_horizon=timedelta(days=10), reference_source=3)
+        >>> # Select the accuracy of beliefs formed 10 days beforehand with respect to the latest belief formed by the first source
+        >>> df.rolling_viewpoint_accuracy(belief_horizon=timedelta(days=10), reference_source=df.lineage.sources[0])
 
         :param belief_horizon: timedelta indicating the belief should be formed at least this duration before knowledge time
         :param belief_horizon_window: optional tuple specifying a horizon window (e.g. between 1 and 2 days before the event value could have been known)
         :param reference_belief_horizon: optional timedelta to indicate that the accuracy should be determined with respect to the latest belief at this duration past knowledge time
         :param reference_source: optional BeliefSource to indicate that the accuracy should be determined with respect to the beliefs held by the given source
-        :param keep_reference_observation: Set to True to return the reference observation used to calculate mape and wape as a DataFrame column
+        :param df_observation: precomputed reference beliefs
```
nhoening: I feel this can be documented a bit more.

1. First, the term "observation" is synonymous with "reference" somehow to me here, can we make the connection between these concepts more clear?
2. This is also a BeliefsDataFrame, right?
3. I see that if `df_observations` is not passed in, we take the latest beliefs (on sefl) as observation. Let's mention that.
4. And I also see that `reference_source` is only used if we did *not* pass in `df_observations`. Mantion that or apply it in either case.
5. Why don't we have this parameter in `fixed_viewpoint_accuracy`, as well?

timely_beliefs/beliefs/classes.py
```diff
@@ -788,26 +857,102 @@ def rolling_viewpoint_accuracy(
...
+            show_accuracy=show_accuracy,
+            active_fixed_viewpoint_selector=active_fixed_viewpoint_selector,
+            reference_source=reference_source,
+            intuitive_forecast_horizon=intuitive_forecast_horizon,
+            interpolate=interpolate,
+        )
+
+    def set_reference_values(
+        self, reference_source: BeliefSource = None
+    ) -> "BeliefsDataFrame":
```
nhoening: This function deserves a docstring (what does it do?) Also, it seems it might be used in `fixed_viewpoint_accuracy` or `rolling_viewpoint_accuracy`? (I might be wrong, but it looks similar to some code in there)

timely_beliefs/beliefs/probabilistic_utils.py
```diff
@@ -401,17 +401,25 @@ def set_truth(
 
 def calculate_crps(df: "classes.BeliefsDataFrame") -> "classes.BeliefsDataFrame":
     """Compute the continuous ranked probability score for a BeliefsDataFrame with a probabilistic (or deterministic)
-    forecast and observation. This function supports a probabilistic observation, too."""
+    forecast (event_value column) and observation (reference_value column).
+    This function supports a probabilistic observation, too.
+
+    References
+    ----------
+    Hans Hersbach. Decomposition of the Continuous Ranked Probability Score for Ensemble Prediction Systems
```
nhoening: It would be neat if we can list literature we use in this package in the main Readme (as well), to show prominently that we use/build on scientific knowledge.

timely_beliefs/visualization/graphs.py
```diff
@@ -0,0 +1,283 @@
...
+                ),
+                alt.Tooltip("source", title="Source"),
+            ],
+        )
+        .properties(width=550, height=200)
+    )
+
+
+def time_series_chart(
+    base,
```
nhoening: What type is base? An `alt.Chart`, right?

timely_beliefs/visualization/graphs.py
```diff
@@ -0,0 +1,283 @@
...
+                    title="Belief horizon (%s)" % belief_horizon_unit,
+                ),
+                alt.Tooltip("source", title="Source"),
+            ],
+        )
+        .properties(width=550, height=200)
+    )
+
+
+def time_series_chart(
```
nhoening: This is also a base chart of kinds, but I find the name "time_series_base_chart" would be more fitting for the "base" chart you create with above method, as it makes sure it has a time-bases x axis and proper tooltips for times.

This one, seems to be responsible for adding actual graphs (or "values") for being displayed in the base time series chart.
So, as usual, I'd propose to rename these functions accordingly.

timely_beliefs/visualization/graphs.py
```diff
@@ -0,0 +1,283 @@
...
+        ts_line_reference_chart = base.mark_rule().encode(x2=alt.X2("event_end:T"))
+    ts_line_reference_chart = ts_line_reference_chart.encode(
+        y=alt.Y(
+            "reference_value",
+            scale=alt.Scale(domain=(event_value_range[0], event_value_range[-1])),
+        ),
+        color=alt.ColorValue("black"),
+        tooltip=[
+            alt.Tooltip(
+                "event_start:T",
```
nhoening: You seem to repeat some tooltips (event_start, event_end) which the `base_chart` function already provides?

timely_beliefs/visualization/selectors.py
```diff
@@ -0,0 +1,195 @@
...
+def horizon_selector(
+    base,
+    horizon_selection_brush: alt.MultiSelection,
+    belief_horizon_unit: str,
+    intuitive_forecast_horizon: bool,
+    unique_belief_horizons,
+) -> alt.LayerChart:
+    bar_chart = (
+        base.mark_rule(orient="vertical")
+        # # Remove the most recent belief horizon from the selector
```
nhoening: Can these lines go?

timely_beliefs/visualization/graphs.py
Flix6x: I tried to add to the list of previously defined tooltips, but it seems as if we need to overwrite the whole thing.

timely_beliefs/visualization/selectors.py
Flix6x: Yes, they just reminded me to finish moving from belief horizon to forecast horizon.

timely_beliefs/beliefs/classes.py
Flix6x: Thanks. I also started renaming `df` to `bdf` so that any difference between a regular pandas DataFrame and a timely-beliefs BeliefsDataFrame will read more clearly.

timely_beliefs/beliefs/classes.py
Flix6x: I was hoping to refactor this code cleanly, but it proved to be harder than it looked. I'll add comments on what the future should hold for this function, but the actual refactoring will be a separate ticket.

Flix6x: Missing support for cleanly concatenating BeliefsDataFrames without loss of metadata was what was holding this back. [This is now fixed.](https://github.com/SeitaBV/timely-beliefs/commit/5f06f89848e42f400b0462ae22c4491c57a8a55b) I then went ahead and did [the actual refactoring[(https://github.com/SeitaBV/timely-beliefs/commit/c266c645970437028596254a62e29deb48881900)

timely_beliefs/beliefs/probabilistic_utils.py
Flix6x: Made this a separate issue. #7

PR #11, opened 2019-09-30
Ridgeline

Flix6x:


timely_beliefs/visualization/utils.py
```diff
@@ -28,7 +32,8 @@ def plot(
     :param ci: The confidence interval to highlight in the time series graph
     :param intuitive_forecast_horizon: If true, horizons are shown with respect to event start rather than knowledge time
     :param interpolate: If True, the time series chart shows a user-friendly interpolated line rather than more accurate stripes indicating average values
-    :param plottable_df: Optionally, specify as plottable DataFrame directly together with a sensor name, a sensor unit and a y-axis value range for the event values (if None, we create it)
+    :param plottable_df: Optionally, specify as plottable DataFrame directly together with a sensor name,
```
nhoening: Is this in place of the bdf param? I mean, are we using the bdf param at all if we pass in plottable_df? If not, we'd need only one argument, which can either be a BeliefsDataFrame, or this tuple of things.

timely_beliefs/visualization/utils.py
```diff
@@ -146,42 +151,41 @@ def plot(
...
 
 
-def timedelta_to_human_range(
-    df: "classes.BeliefsDataFrame"
-) -> Tuple["classes.BeliefsDataFrame", str]:
-    timedelta_span = max(df["belief_horizon"]) - min(df["belief_horizon"])
+def timedelta_to_human_range(s: pd.Series) -> Tuple[pd.Series, str]:
+    """Convert a pandas Series of timedeltas to a pandas Series of floats,
+    and extract a time unit that gives a nice human readable range of floats.
+    """
```
nhoening: Can you give one example in the docstring? It would help to see right away what you mean with "readable range of floats". That term is weird.

timely_beliefs/visualization/utils.py
```diff
@@ -294,3 +300,103 @@ def align_belief_horizons(
...
+def interpret_and_sample_distribution_long_form(
+    df: "classes.BeliefsDataFrame",
+    distribution: str = "uniform",
+    event_value_window: Tuple[float, float] = None,
+) -> pd.DataFrame:
+    """Interpret probabilistic beliefs as a continuous or discrete distribution of possible outcomes.
+    Return a DataFrame with a column for each belief containing probabilities of a given outcome (i.e. the index).
+    """
+    frame = pd.DataFrame()
+    gr = df.for_each_belief()
```
nhoening: This line and the next could be combined.

timely_beliefs/visualization/utils.py
```diff
@@ -294,3 +300,103 @@ def align_belief_horizons(
...
+    return ridgeline_chart
+
+
+def interpret_and_sample_distribution_long_form(
+    df: "classes.BeliefsDataFrame",
+    distribution: str = "uniform",
+    event_value_window: Tuple[float, float] = None,
+) -> pd.DataFrame:
+    """Interpret probabilistic beliefs as a continuous or discrete distribution of possible outcomes.
+    Return a DataFrame with a column for each belief containing probabilities of a given outcome (i.e. the index).
```
nhoening: You're using a function called `drawPDF` to generate this data. Can you explain why and where that function lives?

timely_beliefs/visualization/utils.py
Flix6x: I removed this parameter altogether. It was originally created as a temporary and hacky speed improvement for a demo.

timely_beliefs/visualization/utils.py
Flix6x: I added an example. Is that adequate, or should I comment on how the time unit is chosen?

nhoening: Great!. Well, the docstring says "extract a time unit" and I did not know what that means. Now I see it sets a time unit as a string on the series object.The series' type is now float64, so that "hours" string is set as what  on the series object?

Flix6x: The string is not set on the Series object, but simply the second entry of the returned tuple.

PR #12, opened 2020-07-07
Beliefsdataframe initialisation improvements

Flix6x:


timely_beliefs/beliefs/utils.py
```diff
@@ -313,20 +312,33 @@ def resample_event_start(
...
+    belief_horizon: Union[timedelta, pd.Series],
     cumulative_probability: float = 0.5,
 ) -> List["classes.TimedBelief"]:
     """Turn series entries into TimedBelief objects."""
     beliefs = []
-    for time, value in event_value_series.items():
+    if isinstance(belief_horizon, timedelta):
+        belief_horizon_series = pd.Series(belief_horizon)
+    else:
+        belief_horizon_series = belief_horizon
```
nhoening, line 324: I guess we should check if `belief_horizon_series` end up with the same number of elements as `event_value_series`.

timely_beliefs/beliefs/utils.py
```diff
@@ -313,20 +312,33 @@ def resample_event_start(
...
+    if isinstance(source, BeliefSource):
+        source_series = pd.Series(BeliefSource)
+    else:
+        source_series = source
```
nhoening, line 328: I guess we should check if `source_series` end up with the same number of elements as `event_value_series`.

timely_beliefs/beliefs/utils.py
nhoening: If it's only a timedelta, shouldn't we repeat the value n times (n=-length of event_value_series)?

timely_beliefs/utils.py
```diff
@@ -5,12 +5,12 @@
 import pandas as pd
 
 
-def enforce_utc(dt: datetime) -> pd.Timestamp:
+def enforce_tz(dt: datetime, name: str) -> pd.Timestamp:
```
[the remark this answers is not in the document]
Flix6x: Indeed, this function now serves only to verify a timezone is set.

It seems changing types isn't actually needed anymore. I'm not sure why I made a pandas Timestamp in the first place; maybe it was needed with an earlier version of pandas (I believe I've seen improved support for timezone aware datetimes).

PR #15, opened 2020-07-15
Knowledge horizons take into account updated event resolution after resampling

Flix6x:
This PR addresses what I would consider a bug when resampling. The knowledge time of events was not properly updated after resampling, because it was anchored to the start of an event. In the new default case, knowledge times are anchored to the end of an event.

Consider the following case: when resampling 24 hourly events (each with a knowledge time at the end of the hour) to a single event lasting 24 hours, the knowledge time was still 1 hour after the start of the event. However, the expected outcome is that the new knowledge time is at the end of the new event, so 24 hours after the start of the event. Our new code takes into account the updated event_resolution when determining a knowledge time and/or horizon.


timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -1,25 +1,40 @@
...
 from datetime import datetime, timedelta
 
 from timely_beliefs.sensors.utils import datetime_x_days_ago_at_y_oclock
 
 
+def timedelta_for_end_of_event(
+    event_start: Optional[datetime], event_resolution: timedelta
+) -> Union[timedelta, Tuple[timedelta, timedelta]]:
+    """Knowledge horizon always matches the negative event_resolution.
+    This knowledge horizon is anchored to the event_end.
```
nhoening, line 14: What does it mean if two timedeltas come back?

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -1,25 +1,40 @@
...
 from typing import Optional, Tuple, Union
 from datetime import datetime, timedelta
 
 from timely_beliefs.sensors.utils import datetime_x_days_ago_at_y_oclock
 
 
+def timedelta_for_end_of_event(
+    event_start: Optional[datetime], event_resolution: timedelta
+) -> Union[timedelta, Tuple[timedelta, timedelta]]:
+    """Knowledge horizon always matches the negative event_resolution.
```
nhoening, line 13: Neither the function name nor the docstring really explain (for me) what the function does. Given the event_start and event_resolution, we can compute the knowledge horizon, right?

Maybe the function name needs to mention the knowledge horizon?

timely_beliefs/beliefs/classes.py
```diff
@@ -84,7 +84,8 @@ def __init__(
         elif "belief_time" in kwargs:
             belief_time = tb_utils.enforce_tz(kwargs["belief_time"], "belief_time")
             self.belief_horizon = (
-                self.sensor.knowledge_time(self.event_start) - belief_time
+                self.sensor.knowledge_time(self.event_start, self.event_resolution)
```
nhoening, line 87: I see that the event resolution of a belief trumps the event resolution of a sensor. In other words, the sensor's event resolution is used as fallback, right?

Is that documented? In the Readme, I can't even find information about event resolution of beliefs...

timely_beliefs/sensors/func_store/knowledge_horizons.py
nhoening: I see now the documentation at the top of the module, but developers often don't look there when they check a function. It's still a bit opaque to me what these bounds represent.

timely_beliefs/beliefs/classes.py
Flix6x: The documentation does contain information about even resolutions, but does not yet explain any distinction between the event resolution of the BeliefsDataFrame and the event resolution of the sensor. I find your written questions very useful for writing such new documentation. I'll add it.

Also, you helped me uncover a bug, because `sensor.knowledge_time()` was still sometimes called without specifying the updated event resolution.

timely_beliefs/sensors/func_store/knowledge_horizons.py
Flix6x: Revised accordingly.

timely_beliefs/sensors/func_store/knowledge_horizons.py
Flix6x: I made getting bounds a separate function argument, and clarified all docstrings of these knowledge horizon functions.

Issue #16, opened 2020-07-28
A true mixin approach for subclassing db classes


nhoening:
The current mechanism for subclassing timely-beliefs classes (e.g. `Sensor`) uses [single-table inheritance](https://docs.sqlalchemy.org/en/13/orm/inheritance.html#single-table-inheritance).

However, one drawback of this method is that the table name cannot be chosen independently (e.g. the table needs to be called "sensor"). 

A cleaner way to achieve the primary feature we want to support might be [Mixin classes](https://docs.sqlalchemy.org/en/13/orm/extensions/declarative/mixins.html).

What are these features we should support?

- Users can persist timely-beliefs data to a DB with little work on their own in a new project. This works [in a project of ours](https://github.com/SeitaBV/weatherforecaststorage/blob/3757e6df2011b04820ef8a579ad947ddad986a80/weatherforecast/utils/dbconfig.py#L18). Classes can be made with extra fields and the line `TBBase.metadata.create_all(engine)` creates all tables. This is also documented in the current Readme.
- Users can make their own choices about table names. This is particularly crucial if timely beliefs is used in an existing project.

It seems we could capture current functionality in a true Mixin and still offer the single inheritance approach in a class which uses the mixin but also suggests a table name.

PR #17, opened 2020-08-03
A true mixin approach for subclassing db classes

nhoening:
See #16.

We need a clearer way for the use cases where the user doesn't want to simply use the tables as proposed by timely_beliefs, but has their own table names etc.


.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
```
Flix6x, line 1: We also have a setup.cfg file than has some flake8 settings. I would recommend keeping all these config settings in that file, so possible conflicts (e.g. file length) can be identified within a single file.

.isort.cfg
```diff
@@ -0,0 +1,9 @@
+[settings]
```
Flix6x, line 1: I believe this would become [isort] if moved to our setup.cfg file.

pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[tool.black]
```
Flix6x, line 1: Likely can be moved to config.cfg

timely_beliefs/__init__.py
```diff
@@ -1,12 +1,12 @@
 # flake8: noqa
 
-from timely_beliefs.sensors.classes import Sensor, DBSensor
-from timely_beliefs.sources.classes import BeliefSource, DBBeliefSource
+from timely_beliefs.sensors.classes import DBSensor, Sensor  # isort:skip
+from timely_beliefs.sources.classes import BeliefSource, DBBeliefSource  # isort:skip
```
Flix6x, line 4: Why are these skipped?

timely_beliefs/utils.py
```diff
@@ -21,10 +21,10 @@ def all_of_type(l: Sequence, element_type) -> bool:
 
 
 def replace_multi_index_level(
-    df: "classes.BeliefsDataFrame",  # noqa: F821
+    df: "classes.BeliefsDataFrame",
     level: str,
     index: pd.Index,
-    intersection: bool = False,
+    intersection: bool = False,  # noqa: F821
```
Flix6x, line 27: Is this noqa still needed?

timely_beliefs/sources/classes.py
```diff
@@ -30,39 +33,29 @@ def __lt__(self, other):
...
+    # overwriting name as db field
     name = Column(String(120), nullable=False, default="")
 
     def __init__(self, name: str):
         BeliefSource.__init__(self, name)
-        Base.__init__(self)
 
     def __repr__(self):
-        return "<DBBeliefSource %s (%s)>" % (self.id, self.name)
+        return "<BeliefSource %s (%s)>" % (self.id, self.name)
```
Flix6x, line 49: Why the change?

timely_beliefs/tests/test_db_subclassing.py
```diff
@@ -0,0 +1,127 @@
...
+    assert session.query(RatedSource).filter(RatedSource.rating == 5).count() == 1
+
+    # We made one with default rating (0) and in conftest two are made in advance
+    assert session.query(RatedSource).filter(RatedSource.rating == 0).count() == 3
+
+
+class RatedSourceInCustomTable(Base, BeliefSourceDBMixin):
+    """ A custom db class for representing belief sources.
+    BeliefSource properties are added via the Mixin, so the Base is added by us.
+    Here, we can change the table name.
```
Flix6x, line 45: Great. Please also add some of this information to the main user documentation.

timely_beliefs/beliefs/classes.py
```diff
@@ -127,58 +126,45 @@ def source_id(self):
...
-        if self.__name__ == "DBTimedBelief":
-            return {
-                "polymorphic_on": self.type,
-                "polymorphic_identity": "DBTimedBelief",
-            }
-        else:
-            return {"polymorphic_identity": self.__name__}
+    def sensor_id(cls):
+        return Column(
+            Integer(), ForeignKey("sensor.id", ondelete="CASCADE"), primary_key=True
```
Flix6x, line 157: So when the sensor is deleted, the beliefs are deleted, too? If so, please add that to the docs.

timely_beliefs/beliefs/classes.py
```diff
@@ -127,58 +126,45 @@ def source_id(self):
...
+        )
 
-    def __init__(
-        self, sensor: DBSensor, source: DBBeliefSource, value: float, **kwargs
-    ):
-        TimedBelief.__init__(self, sensor, source, value, **kwargs)
+    @declared_attr
+    def source_id(cls):
+        return Column(Integer, ForeignKey("belief_source.id"), primary_key=True)
```
Flix6x, line 162: So when the source is deleted, the beliefs are not deleted? But the DBTimedeBelief class seems to suggest orphans will be deleted for both sensors AND sources.

timely_beliefs/__init__.py
nhoening: The import order is very important here. The utils need these high-level classes, but isort would put them on top, due some sort order (I believe alphabetical)

timely_beliefs/utils.py
nhoening: I'll try without

timely_beliefs/__init__.py
nhoening: isort has its sorting assumption (I believe alphabetical). In this case, itś crucial that the main classes are there first, as the utils rely on them.

timely_beliefs/sources/classes.py
nhoening: Good catch

timely_beliefs/tests/test_db_subclassing.py
nhoening: I've done so. Please take a look.

.flake8
nhoening: That's true, I'll delete .flake8. For the others, I'm afraid that there is not one way to do everything in one file. 

flake8 can live in pyproject.toml and blakc/isort can go to setup.cfg I believe. I did improve the situation a bit. pyproject.tol is the preferred way forward it seems.

PR #21, opened 2020-08-10
Small fix and usage conveniences for mixin support

Flix6x:
Explicit parameters preferred over kwargs.
Don't init Base in mixin.
Simplify import of mixins.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -53,35 +53,47 @@ class TimedBelief(object):
         sensor: Sensor,
         source: Union[BeliefSource, str, int],
         value: float,
-        **kwargs
+        cumulative_probability: Optional[float] = None,
+        cp: Optional[float] = None,
+        sigma: Optional[float] = None,
+        event_start: Optional[datetime] = None,
+        event_time: Optional[datetime] = None,
+        belief_horizon: Optional[timedelta] = None,
+        belief_time: Optional[datetime] = None,
     ):
         self.sensor = sensor
         self.source = source_utils.ensure_source_exists(source)
         self.event_value = value
 
-        if "cumulative_probability" in kwargs:
-            self.cumulative_probability = kwargs["cumulative_probability"]
-        elif "cp" in kwargs:
-            self.cumulative_probability = kwargs["cp"]
-        elif "sigma" in kwargs:
-            self.cumulative_probability = (
-                1 / 2 + (math.erf(kwargs["sigma"] / 2 ** 0.5)) / 2
-            )
+        if [cumulative_probability, cp, sigma].count(None) not in (2, 3):
+            raise ValueError("Must specify either cumulative_probability, cp, sigma or none of them (0.5 is the default value).")
+        if cumulative_probability is not None:
+            self.cumulative_probability = cumulative_probability
+        elif cp is not None:
+            self.cumulative_probability = cp
+        elif sigma is not None:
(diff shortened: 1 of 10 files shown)


Flix6x:
We haven't release v0.0.10 yet, so we could still include this.

timely_beliefs/beliefs/classes.py
```diff
@@ -162,11 +174,34 @@ def sensor_id(cls):
...
+        sigma: Optional[float] = None,
+        event_start: Optional[datetime] = None,
+        event_time: Optional[datetime] = None,
+        belief_horizon: Optional[timedelta] = None,
+        belief_time: Optional[datetime] = None,
+    ):
         self.sensor_id = sensor.id
         self.source_id = source.id
-        TimedBelief.__init__(self, sensor=sensor, source=source, value=value, **kwargs)
-        Base.__init__(self)
```
Flix6x, line 169: This is actually the small fix: a tb Base is initialized where it shouldn't be. It's a bit hidden within this change list, so that's why this comment is here.

nhoening:
Btw, I'm not sure why we use the this version level. For releases of this size, we could use the second, i.e. 0.1

Am 10. August 2020 22:10:29 MESZ schrieb Felix Claessen <notifications@github.com>:
>We haven't release v0.0.10 yet, so we could still include this.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/SeitaBV/timely-beliefs/pull/21#issuecomment-671563623

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

timely_beliefs/beliefs/classes.py
nhoening: I believe you're correct. This want caught because Base was imported to the file, but I'm not sure what calling Base.__init__ really did in the Mixin. Thanks!

timely_beliefs/tests/test_db_subclassing.py
```diff
@@ -82,9 +82,20 @@ def source_id(cls):
...
+        self,
+        sensor: DBSensor,
+        source: DBBeliefSource,
+        happiness: float = None,
+        **kwargs
+    ):
         self.happiness = happiness
         TimedBeliefDBMixin.__init__(self, sensor, source, **kwargs)
+        [
+            kwargs.pop(key, None)
```
nhoening, line 95: Is this line always need when creating a DB class based on a mixin? I'm not sure I get what this doing so please add a comment.

Flix6x: This is a small experiment in giving developers maximum flexibility. One of the benefits of passing kwargs to a super class is that a developer can change the function signature of the super class without breaking sub classes. In this case there are 2 super classes, which actually have different keyword arguments. I'm assuming no arguments are named the same (we could check that actually), and take out any named arguments used by the first super class to avoid having the second super class complain about keyword arguments it doesn't recognize. This way, the benefit I mentioned works for both super classes.

That said, there are also [popular arguments](https://stackoverflow.com/questions/14626279/inheritance-best-practice-args-kwargs-or-explicitly-specifying-parameters) against passing kwargs to a super class, and for being explicit about the function parameters that each super class needs. I restricted this usage of kwargs to the test only, and for the actual implementation I went with explicit function parameters for maximum documentation. Note that the readme example still uses kwargs (for brevity), but all kwargs go the mixin (and none to `Base`).

I'll add a comment to the test as you suggested.

PR #32, opened 2020-11-02
Set dtypes on empty frame

Flix6x:


timely_beliefs/beliefs/classes.py
```diff
@@ -1415,6 +1429,15 @@ def set_columns_and_indices_for_empty_frame(df, columns, indices):
         ]
     for col in columns + indices:
         df[col] = None
+
+        # Set the correct types
```
nhoening, line 1433: What about `object`? I saw that appear as a type for the "source" field. If it remains untouched with purpose, maybe change the comment to say `Set the pandas types where needed` or so.

nhoening (changes requested):

nhoening (approved):

Issue #34, opened 2020-11-10
BeliefsDataFrame init method from DataFrame changes the input object


Flix6x:
For example:

```
>>> from datetime import timedelta

>>> import pandas as pd
>>> import timely_beliefs as tb


>>> sensor = tb.Sensor(name="some sensor", event_resolution=timedelta(minutes=15))
>>> source_a = tb.BeliefSource("A")
>>> df = pd.DataFrame([[pd.Timestamp("2020-11-03 7:30").tz_localize("utc"), 3]], columns=["event_start", "event_value"])
>>> print(df)
                event_start  event_value
0 2020-11-03 07:30:00+00:00            3
>>> bdf_a = tb.BeliefsDataFrame(df, sensor=sensor, source=source_a, belief_horizon=timedelta(0))
>>> print(df)
                                                                        event_value
event_start               belief_horizon source cumulative_probability             
2020-11-03 07:30:00+00:00 0 days         A      0.5                               3
```

PR #39, opened 2020-11-20
Issue 34 beliefs data frame init method from data frame changes the input object

Flix6x:
I found the same bug with initialization from Series.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -465,6 +465,14 @@ class BeliefsDataFrame(pd.DataFrame):
     ):  # noqa: C901 todo: refactor, e.g. by detecting initialization method
         """Initialise a multi-index DataFrame with beliefs about a unique sensor."""
 
+        # Initialized with a BeliefsSeries or BeliefsDataFrame
+        if len(args) > 0 and isinstance(args[0], (BeliefsSeries, BeliefsDataFrame)):
+            super().__init__(*args, **kwargs)
+            assign_sensor_and_event_resolution(
+                self, args[0].sensor, args[0].event_resolution
+            )
+            return
+
         # Obtain parameters that are specific to our DataFrame subclass
         sensor: Sensor = kwargs.pop("sensor", None)
         event_resolution: timedelta = kwargs.pop("event_resolution", None)
@@ -545,10 +553,10 @@ class BeliefsDataFrame(pd.DataFrame):
             # Interpret initialisation with a pandas Series (preprocessing step of method 3)
             if len(args) > 0 and isinstance(args[0], pd.Series):
                 args = list(args)
+                args[0] = args[0].copy()  # avoid inplace operations
                 args[0] = args[0].to_frame(
                     name="event_value" if not args[0].name else args[0].name
                 )
-                args[0] = args[0].copy()  # avoid inplace operations
                 if isinstance(args[0].index, pd.DatetimeIndex) and event_start is None:
                     args[0].index.name = (
                         "event_start" if not args[0].index.name else args[0].index.name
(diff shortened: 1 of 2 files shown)


Flix6x:
Closes #34

nhoening (approved):
Good find ― an important fix.

nhoening:
I tried to investigate why some tests are failing in this PR, but I couldn't figure it out. Maybe it is more obvious for you.

Flix6x:
> I tried to investigate why some tests are failing in this PR, but I couldn't figure it out. Maybe it is more obvious for you.

## Observation

These tests were failing on an expect column name "reference_value" in tests where accuracy metrics were computed.

## Analysis

The BeliefsDataFrame initialization method from a Series was being triggered, while in fact initializing from a BeliefsSeries. The init method from a Series involves renaming to "event_value" in case the Series doesn't already have a name. The BeliefsSeries actually already had the name "reference_value", but that inexplicably got lost when copying (so then it was renamed and the expected name was not there any more).

## Solution

Moving the copy statement to after the conversion to a frame resolved the issue: at least column names are not lost upon copying. But instead I opted to handle BeliefsDataFrame initialization from a BeliefsSeries or BeliefsDataFrame separately. This leads me to two follow-up tickets:

## Follow-up

- [Refactor BeliefsDataFrame initialization by detecting the initialization method.](https://github.com/SeitaBV/timely-beliefs/issues/40)
- [BeliefsSeries name is lost when copying.](https://github.com/SeitaBV/timely-beliefs/issues/41)

timely_beliefs/beliefs/classes.py
```diff
@@ -465,6 +465,14 @@ def __init__(  # noqa: C901
     ):  # noqa: C901 todo: refactor, e.g. by detecting initialization method
         """Initialise a multi-index DataFrame with beliefs about a unique sensor."""
 
+        # Initialized with a BeliefsSeries or BeliefsDataFrame
+        if len(args) > 0 and isinstance(args[0], (BeliefsSeries, BeliefsDataFrame)):
+            super().__init__(*args, **kwargs)
+            assign_sensor_and_event_resolution(
```
nhoening, line 471: It seems to me that we only call `pd.DataFrame.__init__` here, which won't understand an argument like `beliefs` or `event_start`? These are read in lines 476ff which are now never run for this case.

nhoening (changes requested):

timely_beliefs/beliefs/classes.py
Flix6x: That's correct. Those arguments should only play a role when initializing from a non-(BeliefsSeries or BeliefsDataFrame). That is, all information about events, sensors and sources is already contained in the BeliefsSeries or BeliefsDataFrame.

nhoening: Okay, but how is `args[0]` being preserved within `pd.DataFrame.__init__` Does that do a `self = args[0]` assignment?

Flix6x: The init methods of `pd.Series` and `pd.DataFrame` both try to extract a [block manager](https://uwekorn.com/2020/05/24/the-one-pandas-internal.html), but in different ways. What are you worried about? That the init method can still change the input in this case? Do you propose to add a copy statement here, too?

nhoening: The changing of input was not my worry, but is also interesting to check.

I didn't understand what __init__ is doing with args[0] but apparantly you are confident that the important columns (datetime and event_value) are preserved in the way we expect in a BeliefsDataFrame.

Flix6x: Yes, that I am confident of. I'll add a test for initializing a BeliefsDataFrame from a BeliefsSeres or BeliefsDataFrame.

Issue #35, opened 2020-11-14
Metadata lost on multiplication


Flix6x:
For example, when switching signs:

```
import timely_beliefs as tb


bdf = tb.examples.example_df
bdf.sensor  # <Sensor: Sensor 1>
bdf = bdf * -1
bdf.sensor # None
```

Flix6x:
Workaround for the particular case when needing to switch signs:

```
bdf = -bdf
bdf.sensor  # <Sensor: Sensor 1>
```

Or when you need to switch from negative to positive values:

```
bdf = bdf.abs()
bdf.sensor  # <Sensor: Sensor 1>
```

Flix6x:
Closing this as no longer reproducing.

While reviewing #249, which linked this issue, I checked the snippet above against current `main` on pandas 1.5.3, 2.0.3, 2.1.1, 2.1.4, 2.2.3 and 2.3.3. `bdf * -1` keeps its sensor on all of them, and metadata survives the surrounding operations I tried alongside it (`iloc`, `head`, `copy`, `to_frame`, `groupby.mean`, `unstack`, `resample` and DataFrame slicing).

The `_constructor_from_mgr` and `_constructor_expanddim_from_mgr` methods that are now on `main` appear to have settled this. Sorry it stayed open so long after being fixed in passing.

If you do still see metadata dropped on arithmetic somewhere, please reopen with the pandas version and the expression, and I will take another look.

Issue #26, opened 2020-09-15
Tests fail with pandas==1.1(metadata lost on groupby)


Flix6x:
Until [this issue](https://github.com/pandas-dev/pandas/issues/29442) is fixed, timely-beliefs only supports pandas<1.1.

nhoening:
The pandas fix has 1.1.1 as target, so 1.1.0 wouldn't work. Or they backport the fix.

Am 15. September 2020 20:05:25 MESZ schrieb Felix Claessen <notifications@github.com>:
>Until [this issue](https://github.com/pandas-dev/pandas/issues/29442)
>is fixed, timely-beliefs only supports pandas<1.1.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/SeitaBV/timely-beliefs/issues/26

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

Flix6x:
The fix was released as part of pandas 1.1.5. There was a [backport](https://github.com/pandas-dev/pandas/pull/37628), too, but our relevant tests still fail. @nhoening can you comment if my procedure below is correct? I am probably misunderstanding how backports actually work. Basically, I thought a backport of a fix would lead to that fix being present in the code of relevant older versions as available on pypi, perhaps as a developer release. However, the latest release of 1.1.4 on pypi is on October 30th, while the backport was merged on November 4th.

Here I try to install a previous version of pandas with the backport supposedly in place:

```
pip install pandas==1.1.4 --no-cache --force-reinstall
```

Then I run the test that I would expect to pass (if this focused test fails, other integration tests fail, too, of course):

```
pytest -k test_groupby_retains_metadata
```
This test, and all others, too, are passing against pandas==1.1.5 on #23 .

nhoening:
It was indeed not obvious to find, but they had it set on 1.2 in the beginning of the PR, but then backported it to 1.1.5 when they were done.
You can see this [here](https://github.com/pandas-dev/pandas/pull/37461#pullrequestreview-522002408) (click "show resolved"), where they removed the release note entry from the 1.2 part (comment: "Force pushed to drop this commit").

Issue #22, opened 2020-08-14
Lost metadata when resampling


Flix6x:
[Pandas 1.1.0 fixed a bug](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.1.0.html#groupby-resample-rolling) where [resampling causes the subclassed data structure to be lost](https://github.com/pandas-dev/pandas/issues/28330).

Another effort is undertaken to [preserve metadata when using groupby](https://github.com/pandas-dev/pandas/issues/29442), a fix for which is scheduled for release version 1.1.1.

That means by updating our dependency to pandas=1.1.1 we should be able to simplify (and increase the speed of) some of our resampling methods by not having to reinitialize the BeliefsDataFrame. I already had a test stashed for just this case.

Flix6x:
Release of this fix has been postponed, at least to pandas 1.1.3.

Flix6x:
Fix was released as part of pandas 1.1.5.

PR #43, opened 2020-12-11
Issue 35 metadata lost on multiplication

Flix6x:
This PR adds no additional logic, only a test to check whether Issue #35 is successfully resolved (plus some refactoring of test util functions). Note that this is a merge into resample-while-keeping-metadata, and that branch (with PR #23) actually contains the logic that resolves this issue (as a side effect, because I was actually addressing deeper issues there: #22 and #26).

### DIFF
--- a/timely_beliefs/tests/test_belief_io.py
+++ b/timely_beliefs/tests/test_belief_io.py
@@ -570,3 +570,19 @@ def test_groupby_retains_subclass_attribute(att, args):
     df2 = getattr(df.groupby("x"), att)(*args)
     print(df2)
     assert df2.a == "b"
+
+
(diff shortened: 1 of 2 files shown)


nhoening:
Please add in title or subscription what you are doing here. This is just improving our tests and no addition to any logic right?

Flix6x:
This PR indeed adds no additional logic, only a test to check whether Issue #35  is successfully resolved (plus some refactoring of test util functions). Note that this is a merge into resample-while-keeping-metadata, and _that_ branch (with PR #23) actually contains the logic that resolves this issue (as a side effect, because I was actually addressing deeper issues there: #22 and #26).

nhoening (approved):

PR #44, opened 2021-01-13
More user friendly conversions

Flix6x:
Mostly reindexing from belief_time to belief_horizon, but also getting rid of future warning and one bug fix where selecting the most recent belief can result in selecting a NaN belief, which isn't the one we want to select.

I still have 1 test (test_crps) to fix.

### DIFF
--- a/timely_beliefs/beliefs/utils.py
+++ b/timely_beliefs/beliefs/utils.py
@@ -20,17 +20,27 @@ from timely_beliefs.sources import utils as source_utils
 def select_most_recent_belief(
     df: "classes.BeliefsDataFrame",
 ) -> "classes.BeliefsDataFrame":
-    """Drop all but most recent belief."""
+    """Drop all but most recent non-NaN belief."""
     if "belief_horizon" in df.index.names:
-        return df.groupby(level=["event_start", "source"], group_keys=False).apply(
-            lambda x: x.xs(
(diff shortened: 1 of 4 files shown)


timely_beliefs/beliefs/utils.py
```diff
@@ -445,10 +445,13 @@ def set_reference(
     if return_expected_value is True:
         reference_df = reference_df.for_each_belief(get_expected_belief)
 
+    belief_timing_col = (
+        "belief_time" if "belief_time" in reference_df.index.names else "belief_horizon"
```
nhoening, line 459: Not sure I understand exactly what is happening here. But "belief_horizon" is used as a name in the concatenated and returned dataframe (lines 458ff). Is that still accurate when we dropped it?

nhoening (approved):

timely_beliefs/beliefs/utils.py
Flix6x: It is. We first get rid of the belief timing column and then reconstruct our own, where we choose the horizon perspective explicitly.

Flix6x:
[Get rid of future_warning](https://trello.com/c/aS9Yt3ef/66-get-rid-of-futurewarning)

Flix6x:
[Convert belief timing index level in more places](https://trello.com/c/kqYsSvGF/64-convert-belief-timing-index-level-in-more-places)

PR #45, opened 2021-01-14
More user friendly initialization

Flix6x:
I added a long overdue documentation section on different ways to initialize a BeliefsDataFrame, which was mostly still hidden in 
its class docstring, and included the shortest possible way near the top of the main Readme, too. I also took the opportunity to 1) build in some flexibility to specify durations and timestamps in different formats, 2) rename a variable, and 3) add a shorthand for a TimedBelief, just Belief seems nice.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -432,8 +432,7 @@ class BeliefsDataFrame(pd.DataFrame):
         Method 2:   pass a pandas DataFrame with columns ["event_start", "belief_time", "source", "event_value"]
                     - Optional column: "cumulative_probability" (the default is 0.5)
                     - Alternatively, use keyword arguments to replace columns containing unique values for each belief
-        Method 3:   pass a pandas Series with DatetimeIndex and keyword arguments for "source"
-                    - Optional keyword argument: "belief_horizon" (the default is sensor.event_resolution)
+        Method 3:   pass a pandas Series with DatetimeIndex and keyword arguments for "belief_time" or "belief_horizon", and "source"
                     - Alternatively, use the "event_start" keyword argument to ignore the index
 
     In addition to the standard DataFrame constructor arguments,
@@ -1526,3 +1525,6 @@ def downsample_beliefs_data_frame(
         ],
         axis=1,
     ).set_index([belief_timing_col, "source", "cumulative_probability"], append=True)
+
+
+Belief = TimedBelief
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -23,6 +23,8 @@ from timely_beliefs.sources.classes import BeliefSource, DBBeliefSource
 from timely_beliefs.visualization import utils as visualization_utils
 
 METADATA = ["sensor", "event_resolution"]
+DatetimeLike = Union[datetime, str, pd.Timestamp]
+TimedeltaLike = Union[timedelta, str, pd.Timedelta]
 
 
 class TimedBelief(object):
@@ -57,10 +59,10 @@ class TimedBelief(object):
         cumulative_probability: Optional[float] = None,
         cp: Optional[float] = None,
         sigma: Optional[float] = None,
-        event_start: Optional[datetime] = None,
-        event_time: Optional[datetime] = None,
-        belief_horizon: Optional[timedelta] = None,
-        belief_time: Optional[datetime] = None,
+        event_start: Optional[DatetimeLike] = None,
+        event_time: Optional[DatetimeLike] = None,
+        belief_horizon: Optional[TimedeltaLike] = None,
+        belief_time: Optional[DatetimeLike] = None,
     ):
         self.sensor = sensor
         self.source = source_utils.ensure_source_exists(source)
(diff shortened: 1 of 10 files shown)


Flix6x:
[Document BeliefsDataFrame initialisation](https://trello.com/c/3DK4qUTs/60-document-beliefsdataframe-initialisation)

timely_beliefs/docs/init.md
```diff
@@ -0,0 +1,91 @@
...
+    >>> bdf = tb.BeliefsDataFrame(df, belief_horizon=timedelta(hours=0), source=source, sensor=sensor)
+    >>> print(bdf)
+                                                                            event_value
+    event_start               belief_horizon source cumulative_probability             
+    2000-01-03 09:00:00+00:00 0 days         EPEX   0.5                              63
+    2000-01-03 10:00:00+00:00 0 days         EPEX   0.5                              60
+
+## From a CSV file
+
+The utility function `tb.read_csv` lets you load a BeliefsDataFrame from a csv file (e.g. example/temperature.csv).
```
nhoening, line 66: So `example/temperature.csv` lets me see a working example? Tell me explicitly to go and look there, as this paragraph would not be sufficient for me to implement this way of loading data.

timely_beliefs/docs/init.md
```diff
@@ -0,0 +1,91 @@
...
+To write a BeliefsDataFrame to a csv file, just use the pandas way:
+
+    >>> bdf.to_csv("data.csv")
+    >>> tb.read_csv("data.csv", source=source, sensor=sensor)
+
+## From a list of TimedBeliefs
+
+Create a list of `TimedBelief` or `DBTimedBelief` objects and use it to initialize a BeliefsDataFrame.
+
+_This method is meant for applications with a database._
```
nhoening, line 79: Why is that? This example seems to work well without a DB. Do you mean that the other methods are otherwise preferrable?

timely_beliefs/beliefs/classes.py
```diff
@@ -52,18 +54,31 @@ def __init__(
         self,
         sensor: Sensor,
         source: Union[BeliefSource, str, int],
-        value: float,
+        event_value: Optional[float] = None,
```
nhoening, line 57: I'd add a comment here that `value` and `event_value` are the same with just different names. Really, I think I'd like to keep one parameter here. If you want to move towards `event_value`, we should do that whole-hearted. Maybe deprecate `value`?

nhoening (changes requested):

timely_beliefs/beliefs/classes.py
```diff
@@ -1498,3 +1529,6 @@ def downsample_beliefs_data_frame(
         ],
         axis=1,
     ).set_index([belief_timing_col, "source", "cumulative_probability"], append=True)
+
+
+Belief = TimedBelief
```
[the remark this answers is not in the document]
Flix6x: It's true that this doesn't really add functionality. I just added it so that examples in the documentation become a bit shorter.

If we decide **for** it, I would argue to do at least the same for `Source = BeliefSource`, as one of our three custom defined core objects `Sensor`, `Source` and `Belief`. Would it make sense to also have `DataFrame = BeliefsDataFrame`? Although at first glance that seems confusing, it's type would still be BeliefsDataFrame and it would arguably look good in examples. Our first example in the main readme would then become:

```
>>> import timely_beliefs as tb
>>> df = tb.DataFrame([tb.Belief(tb.Sensor("Indoor temperature", "°C"), tb.Source("Thermometer"), 21, event_time="2000-03-05 11:00Z", belief_horizon="0H")])
```

This may be a slippery slope into renaming a lot of things, though. But I would argue not to go further than this. For example, no shorthands for other classes like `[...]DBMixin` classes or other named variables like `belief_horizon`.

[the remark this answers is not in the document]
Flix6x: For clarity, the renaming of "value" to "event_value" in the TimedBelief initialization in this PR is not an attempt to allow both a shorthand and the full variable name. Rather, that is an attempt to migrate away from inconsistent variable naming. But instead of immediately replacing the variable name, I wanted to try out writing a user warning for an upcoming deprecation.

timely_beliefs/docs/init.md
Flix6x: You're right, this is an acceptable method also without a database. I'll remove the comment.

timely_beliefs/beliefs/classes.py
Flix6x: This is an attempt to migrate away from inconsistent variable naming. But instead of immediately replacing the variable name, I wanted to try out writing a user warning for an upcoming deprecation. I'd like to deprecate it in a later (major) release.

timely_beliefs/beliefs/classes.py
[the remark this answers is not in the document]
nhoening: Yes, a deprecation across two months or so sounds like the right approach. Even though there are probably no users we don't talk to in person right now.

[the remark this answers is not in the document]
nhoening: I can't decide pro-shorthand right now. It seems like the gains (short examples) are small, while the potential downsides (confusion, mixing with other libraries' namespaces) seem potentially large.

Maybe postpone this specific discussion? It's larger than this PR.

nhoening (approved):

PR #50, opened 2021-02-08
Form new beliefs given some model

Flix6x:
This introduces a new BeliefsDataFrame method `form_beliefs` that creates new beliefs given a `model`, a `belief_time` and an `event_time_window`.

Much credit goes to the AI Master students from the University of Amsterdam that also helped us implement [ridgeline plots for probabilistic forecasts](https://github.com/SeitaBV/timely-beliefs/blob/master/timely_beliefs/docs/viz.md).


Flix6x:
Test setup is failing for Python 3.7, with `RuntimeError: Scikit-learn requires Python 3.8 or later. The current Python version is 3.7.13`. My proposal would be to make sktime (which is requiring scikit-learn) a soft dependency, installable with `pip install timely-beliefs[forecast]` (after merging #107), and exclude tests relating to forming new beliefs for python==3.7.

timely_beliefs/tests/test_belief_formation.py
```diff
@@ -0,0 +1,41 @@
...
+        source=BeliefSource("Source C"),
+        event_start=datetime(2000, 1, 3, 13, tzinfo=pytz.utc),
+        forecaster=forecaster,
+    )
+
+    # Operation should not affect original DataFrame
+    pd.testing.assert_frame_equal(df_in, df_in_copy)
+
+    assert df_out.values[0] == forecast
+    df = pd.concat([df, df_out])
```
nhoening, line 40: I believe these two last lines can go?

timely_beliefs/beliefs/classes.py
```diff
@@ -1467,6 +1492,124 @@ def resample_events(
 
         return df
 
+    @hybrid_method
+    def form_beliefs(
```
nhoening, line 1496: Why not "forecast_beliefs"? 
We pass in a "forecaster", so that's clearly what is happening, and we can be more specific.

timely_beliefs/beliefs/classes.py
```diff
@@ -1467,6 +1492,124 @@ def resample_events(
...
+            raise NotImplementedError("Backcasting is not implemented.")
+        if self.empty:
+            # At least 1 belief
+            raise NotImplementedError(
+                "Cannot form new beliefs given an empty data frame."
+            )
+        if self.lineage.number_of_sources != 1:
+            # Single source only
+            raise NotImplementedError(
+                "Cannot form new beliefs given multi-sourced data. Consider picking 1 source, e.g. using: df = df[df.index.get_level_values('source') == df.lineage.sources[0]]"
```
nhoening, line 1545: Why is this not supported? Technical reasons?
Also, we could accept a source to narrow down what data to use first, right?

timely_beliefs/beliefs/classes.py
```diff
@@ -1467,6 +1492,124 @@ def resample_events(
...
+            )
+        if self.lineage.probabilistic_depth != 1:
+            # Deterministic beliefs only
+            raise NotImplementedError(
+                "Cannot form new beliefs given probabilistic data. Consider pre-computing deterministic beliefs, e.g. using: df = df.make_deterministic()"
+            )
+        if self.lineage.number_of_beliefs != self.lineage.number_of_events:
+            # One belief per event
+            raise NotImplementedError(
+                "Cannot form new beliefs given multiple beliefs about the same event. Consider picking a single belief per event, e.g. using: df = belief_utils.select_most_recent_belief(df)"
```
nhoening, line 1555: Is "df.rolling_viewpoint()" another possibility to mention here?

timely_beliefs/beliefs/classes.py
```diff
@@ -1467,6 +1492,124 @@ def resample_events(
...
+        event_time_window = tuple(
+            dt.astimezone(utc).replace(tzinfo=None) for dt in event_time_window
+        )
+        df.index = df.index.tz_convert(utc).tz_localize(None)
+
+        # Resample to the resolution the data already has, just to set the index frequency, which sktime expects
+        df = df.reindex(
+            pd.date_range(
+                df.index[0],
+                event_time_window[0],
```
nhoening, line 1580: I expected event_time_window[-1] here. Or maybe belief_time. Can you motivate this?

timely_beliefs/beliefs/classes.py
```diff
@@ -1467,6 +1492,124 @@ def resample_events(
...
+
+        # Apply model
+        if isinstance(forecaster, BaseForecaster):
+            forecast_event_starts = pd.date_range(
+                event_time_window[0],
+                event_time_window[1],
+                freq=df.event_resolution,
+                closed="left",
+            )
+            forecaster.fit(df["event_value"])
```
nhoening, line 1597: At this time, our BeliefDataFrames do not support regressors to be used here, correct? For example, forecasting wind speed based on time of day and temperature - how would we do that? (maybe another ticket)

nhoening (commented):
Awesome to finally bring this in, and to start usign sktime!

I have mainly questions to offer.

timely_beliefs/tests/test_belief_formation.py
Flix6x: Thanks for noticing. I'm expanding the test to check this concatenation, too. It turns out that my `inplace` flag was a misnomer.

timely_beliefs/beliefs/classes.py
Flix6x: I'm adding this motivation:

The new index ends just before the start of the first forecast (which may introduce trailing NaN values), as sktime expects no gap between the indices of the input and forecasts when applying seasonal periodicity.

timely_beliefs/beliefs/classes.py
Flix6x: Good idea to separate this into another ticket. It can be done by passing an extra `X` to the fit function, for example, like this:
```
X=df.loc[:, df.columns != "event_value"]
```
Here, I'm assuming the external regressors are given as additional columns in the BeliefsDataFrame.

Testing it with sktime's naive forecaster will just lead to the forecaster ignoring `X` though, so that doesn't feel like a wholesome test.

timely_beliefs/beliefs/classes.py
Flix6x: > Why is this not supported? Technical reasons?

Missing implementation, for which I'm still unsure which direction to take it. If two sources have different expectations about the last value, what should the forecaster return? We don't yet have a feature to compute aggregate beliefs. For example, consider this input data:
```
                                                                                     event_value
event_start               belief_time               source   cumulative_probability             
2000-01-03 09:00:00+00:00 2000-01-01 01:00:00+00:00 Source A 0.5                           100.0
                                                    Source B 0.5                            25.0
```
Given equal weights, the input data could become:
```
                                                                                       event_value
event_start               belief_time               source     cumulative_probability             
2000-01-03 09:00:00+00:00 2000-01-01 01:00:00+00:00 Source A+B 0.5                            62.5
```
and the forecast could be `62.5` with `Source f(A+B)`. Alternatively, the forecaster could return:
- `[25.0, 100.0]` with separate sources `[Source f(A), Source f(B)]`.
- `[25.0, 62.5, 100.0]` with `Source f(A+B, d_A, d_B)` and CP values depending on assumptions on the underlying distributions `d` (e.g. CP values of `[0.5, 0.5, 0.5]`, `[0.33, 0.5, 0.67]` or `[0.25, 0.5, 0.75]`).
- `[25.0, 80.0, 100.0]` with `Source f(A+B, w)`given some skewed weight distribution `w` over the sources that favours source A.

> Also, we could accept a source to narrow down what data to use first, right?

We could. But imo that should not be the task of this function.

nhoening: >  But imo that should not be the task of this function.

Agreed.

timely_beliefs/beliefs/classes.py
Flix6x: This will get pretty meta, but a forecast **is** a belief, and a belief is not something you typically forecast, but rather is something you form.

Unless I forecast that you will believe I am mistaken, in which case my belief is the forecast and your belief is the event my belief is about.
```
                                                                                   event_value  sensor
event_start               belief_time               source cumulative_probability             
2022-09-16 13:20:00+02:00 2022-09-16 15:10:00+02:00 Felix  0.5                           False Nicolas
2022-09-16 13:25:00+02:00 2022-09-16 15:10:00+02:00 Felix  0.5                            True Nicolas
```

:)

timely_beliefs/beliefs/classes.py
nhoening: Did we make an issue out of this yet? I can't find one.

nhoening: Ah, I see you already add the `X` in this PR. I believe a little comment would help.

And is there a case where columns in X would be bad to include?

We could also add ab optional parameter to `form_beliefs`, e.g. "X_column_names".

nhoening (commented):
I have one comment/question about X.

timely_beliefs/beliefs/classes.py
Flix6x: I created https://github.com/SeitaBV/timely-beliefs/issues/116 just now.

nhoening (approved):

PR #53, opened 2021-03-29
Support timed belief db mixin

Flix6x:
Several improvements to the TimedBeliefDBMixin, including:

- New method to persist TimedBelief objects from a BeliefsDataFrame
- Set default table indices
- Set default cumulative probability
- Rename query method to prevent clashing with SQLAlchemy

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -269,7 +269,7 @@ class TimedBeliefDBMixin(TimedBelief):
         event_not_before: Optional[datetime] = None,
         belief_before: Optional[datetime] = None,
         belief_not_before: Optional[datetime] = None,
-        source: Optional[Union[int, List[int], str, List[str]]] = None,
+        source: Optional[Union[BeliefSource, List[BeliefSource]]] = None,
     ) -> "BeliefsDataFrame":
         """Search a database session for beliefs about sensor events.
         :param session: the database session to use
@@ -278,7 +278,7 @@ class TimedBeliefDBMixin(TimedBelief):
         :param event_not_before: only return beliefs about events that start after this datetime (inclusive)
         :param belief_before: only return beliefs formed before this datetime (inclusive)
         :param belief_not_before: only return beliefs formed after this datetime (inclusive)
-        :param source: only return beliefs formed by the given source or list of sources (pass their id or name)
+        :param source: only return beliefs formed by the given source or list of sources
         :returns: a multi-index DataFrame with all relevant beliefs
 
         TODO: rename params for clarity: event_finished_before, event_starts_not_before (or similar), same for beliefs
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -230,18 +230,18 @@ class TimedBeliefDBMixin(TimedBelief):
         )
 
     @classmethod
-    def add_all(
+    def add_to_session(
         cls,
         session: Session,
         beliefs_data_frame: "BeliefsDataFrame",
         commit_transaction: bool = False,
     ):
-        """Add a BeliefsDataFrame as timed beliefs to the database session.
+        """Add a BeliefsDataFrame as timed beliefs to a database session.
 
         :param session: the database session to use
         :param beliefs_data_frame: the BeliefsDataFrame to be persisted
         :param commit_transaction: if True, the session is committed
-                                   if False, you can add still other data to the session
+                                   if False, you can still add other data to the session
                                    and commit it all within an atomic transaction
         """
         # Belief timing is stored as the belief horizon rather than as the belief time
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -75,7 +75,7 @@ class TimedBelief(object):
         elif event_value is not None:
             self.event_value = event_value
         else:
-            # todo: deprecate the 'value' argument in favor of 'event_value'
+            # todo: deprecate the 'value' argument in favor of 'event_value' (announced v1.1.0)
             import warnings
 
             warnings.warn(
@@ -259,7 +259,7 @@ class TimedBeliefDBMixin(TimedBelief):
     @tb_utils.append_doc_of("TimedBeliefDBMixin.query_all")
     def query(cls, *args, **kwargs):
         """Function will be deprecated. Please switch to using query_all."""
-        # todo: deprecate this function, which can clash with SQLAlchemy's Model.query()
+        # todo: deprecate this function (announced v1.3.0), which can clash with SQLAlchemy's Model.query()
         import warnings
 
         warnings.warn(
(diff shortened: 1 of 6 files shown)


timely_beliefs/beliefs/classes.py
```diff
@@ -221,15 +230,56 @@ def __init__(
...
+        :param commit_transaction: if False, you can add still other data to the session
+                                   and commit it all within an atomic transaction
+        """
+        belief_records = (
+            beliefs_data_frame.convert_index_from_belief_time_to_horizon()
+            .reset_index()
+            .to_dict("records")
+        )
+        beliefs = [
+            cls(**dict(**d, sensor=beliefs_data_frame.sensor)) for d in belief_records
```
nhoening, line 252: I believe you don't need the extra `**dict` wrapping here:

```
>>> class A(object):
. . .       def __init__(self, a=2, b=3, c=4, d=5):
. . .            self.a = a
. . .            self.b = b
. . .            self.c = c
. . .            self.d = d

>>> d1 = dict(a=1, b=2, c=3)
>>> a = A(**d1)
>>> f"A(a={a.a} b={a.b} c={a.c} d={a.d})"
'A(a=1 b=2 c=3 d=5)'
>>> a = A(**d1, d=4)
>>> f"A(a={a.a} b={a.b} c={a.c} d={a.d})"
'A(a=1 b=2 c=3 d=4)'
```

timely_beliefs/beliefs/classes.py
```diff
@@ -221,15 +230,56 @@ def __init__(
...
+        cls,
+        session: Session,
+        beliefs_data_frame: "BeliefsDataFrame",
+        commit_transaction: bool = False,
+    ):
+        """Add a BeliefsDataFrame as timed beliefs to the database session.
+
+        :param session: the database session to use
+        :param beliefs_data_frame: the BeliefsDataFrame to be persisted
+        :param commit_transaction: if False, you can add still other data to the session
```
nhoening, line 243: It's probably clearer to say what happens if you set this to `True`.

timely_beliefs/beliefs/classes.py
```diff
@@ -221,15 +230,56 @@ def __init__(
...
+                                   and commit it all within an atomic transaction
+        """
+        belief_records = (
+            beliefs_data_frame.convert_index_from_belief_time_to_horizon()
```
nhoening, line 247: Maybe a short comment why this conversion has to be done?

timely_beliefs/beliefs/classes.py
```diff
@@ -239,6 +289,8 @@ def query(
         :param belief_before: only return beliefs formed before this datetime (inclusive)
         :param belief_not_before: only return beliefs formed after this datetime (inclusive)
         :param source: only return beliefs formed by the given source or list of sources (pass their id or name)
+        :param sensor_cls: set to your own DBSensor class implementation if it uses a different table name
```
nhoening, line 292: Can't we get the sensor class from `sensor.__class__`? (and the same for `source.__class__`)

timely_beliefs/beliefs/classes.py
```diff
@@ -221,15 +230,56 @@ def __init__(
...
+        import warnings
+
+        warnings.warn(
+            "Function 'query' will be replaced by 'query_all'.",
+            FutureWarning,
+        )
+        return cls.query_all(*args, **kwargs)
+
+    @classmethod
+    def query_all(
```
nhoening, line 272: Maybe this is personal taste, but to me the name "query_all" is a bit misleading ― we don't get all the beliefs, we query the ones from a sensor and can filter down by times and source. Maybe "query_by_sensor" is describing better what this does?

Not straightforward, that's for sure.

timely_beliefs/tests/test_db_subclassing.py
```diff
@@ -136,3 +136,6 @@ def test_custom_source_and_beliefs_with_mixin(db):
 
     assert "my_belief_source" in db.tables.keys()
     assert "my_timed_belief" in db.tables.keys()
+
+    bdf = JoyfulBeliefInCustomTable.query(session, sensor)
```
nhoening, line 140: Is this supposed to be `query_all`? If so, why does this assertion work? Is `bdf` maybe an `sqlachemy.Query` object instead of a `BeliefsDataFrame`?

nhoening (changes requested):
Good stuff!
I have only smallish comments, only two I really think need to be done.

timely_beliefs/beliefs/classes.py
nhoening: Just saw that you don't expect the sources as an object (why not?), so yes then you need to ask which type the objects really should have.

Also, that you're typing the `sensor` parameter as `DBSensor`, which is typically not used by libraries implementing their sensor using the `SensorDBMixin`. Maybe that is a better expected type here...

timely_beliefs/tests/test_db_subclassing.py
Flix6x: I used `query` on purpose, to have a test that triggers the corresponding FutureWarning. I've been following this strategy for all deprecations, for two reasons:

1. The test suite will show FutureWarnings for upcoming deprecations, which keeps reminding developers to actually deprecate at some point.
2. The developer that implements the actual deprecation will break the test, and is therefore made aware of the fact that dependent code may break (and hopefully, reflect on whether enough time has passed since the deprecation was announced).

I'm glad you noticed this, because giving it some more thought I can come up with a few things to improve this strategy:

- Add a todo to the assert statement to amend the upcoming deprecation.
- Note the version at which the deprecation was announced.

timely_beliefs/beliefs/classes.py
Flix6x: That's a very neat solution, thanks.

timely_beliefs/beliefs/classes.py
```diff
@@ -303,23 +343,9 @@ def query(
...
-                raise ValueError(
-                    "Query by source failed: query only possible by integer id or string name. Failed sources: %s"
-                    % unidentifiable_list
-                )
-            else:
-                q = q.join(DBBeliefSource).filter(
-                    (cls.source_id.in_(id_list)) | (DBBeliefSource.name.in_(name_list))
-                )
+            sources: list = [source] if not isinstance(source, list) else source
+            source_cls = sources[0].__class__
```
nhoening, line 347: The `source` parameter to this function is of type `Optional[Union[int, List[int], str, List[str]]]`, so I don't think the `__class__` you get here will be anything else than `int` or `str`.

nhoening (commented):
Almost there! One comment/question I have is about the `source` parameter in `search_session`. Don't we still need a `source_cls` parameter?

Alternatively, pass sources as objects. Your code (lines 347f) seems to already assume they are.

timely_beliefs/beliefs/classes.py
Flix6x: Thanks, I just forgot to update the type annotation and docstring. The `__class__` should give the BeliefSource, or whatever subclass is used (like DataSource in FlexMeasures).

Flix6x:
Does my last commit resolve you last question?

nhoening (approved):

PR #55, opened 2021-04-11
Update altair

Flix6x:
We're now also taking advantage of the new labelExpr feature in vega-lite to get less crowded y-axis labels on ridgeline plots.

As we're updating a major dependency, we also updated dev requirements and the make targets around them. Furthermore, we changed our CI, so that it doesn't use the pinned dev dependencies, but rather simulates an actual install.

### DIFF
--- a/timely_beliefs/visualization/graphs.py
+++ b/timely_beliefs/visualization/graphs.py
@@ -340,8 +340,11 @@ def probabilistic_chart(
         stroke=alt.condition(
             selectors.ridgeline_hover_brush, alt.value("black"), alt.value("lightgray")
         ),
-        strokeWidth=alt.condition(
-            selectors.ridgeline_hover_brush, alt.value(2.5), alt.value(0.5)
+        strokeWidth=alt.StrokeWidthValue(
+            0.5,
+            condition=alt.StrokeWidthValue(
+                2.5, selection=selectors.ridgeline_hover_brush.name
+            ),
         ),
     )
     area_chart = base_chart.mark_area(interpolate="monotone", fillOpacity=0.6).encode(
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
 services:
   - postgresql
 install:
+  - pip install --upgrade pip  # latest resolver
   - python setup.py install  # testing a non-pinned install, suitable for a library
   - pip install pytest
   - pip install .
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,9 @@ python:
 services:
   - postgresql
 install:
-  - pip install --upgrade pip  # latest resolver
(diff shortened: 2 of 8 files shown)


nhoening:
How should I review this? My Altair experience is very small.

Flix6x:
Good point. Try upgrading Altair and making a ridgeline plot and an accuracy dashboard (see examples with the included temperature dataset in the documentation).

Maybe I can also try initializing the belief time to the middle of its range, so that you immediately see a fixed viewpoint forecast instead of having to select a belief time first.

nhoening:
Just a note about visualisation docs: docs/viz.md says "This will create an interactive Vega-Lite chart (Satyanarayan et al., 2016) like the one in the screenshot at the top of this Readme." I believe that is outdated since the Readmes were restructured and now is a good time to fix this.

nhoening (approved):
I added a commit and have one comment on docs.

nhoening:
As part of my dev/requirements.txt update, the Pandas version is no `1.2.3`. As a consequence of that, we stopped supporting 3.6 here. 

@Flix6x please advise which version we should & want support.

Flix6x:
I'd be fine with having timely beliefs stop supporting python==3.6, but I don't understand why this would be needed.

I would expect that the requirements.txt in timely-beliefs doesn't force flexmeasures into using that exact same version of pandas. After all, the tb setup.py is more flexible regarding what version you should use (`pandas>=1.1.5`).

nhoening:
If no other packages limit the requirements, `pandas>=1.1.5` will pick the latest stable version which is now `1.2.3`. Having Python3.6 installed doesn't tell your system which Pandas version you should install I believe.

That is one problem (Python3.6 users run into a bad dependency), which can in principle be solved locally but is tough to document well.

Another problem is what came up here: Our CI builds. To support Python3.6 there, we would have to add `<1.2` to the requirements. And that has other maintenance issues. The easy way out of this one is to stop running the tests against 3.6 (removing one line in `.travis.yaml`).

timely_beliefs/visualization/graphs.py
```diff
@@ -259,7 +259,7 @@ def source_vs_hour_chart(
                 "event_start:O",
                 timeUnit="hours",
                 axis=alt.Axis(domain=False, ticks=False, labelAngle=0),
-                scale=alt.Scale(domain=list(range(24))),
+                # scale=alt.Scale(domain=list(range(24))),
```
nhoening, line 262: Why is this commented out? Do we want to keep it?

Flix6x: I had hoped to fix it, but was unable to. I believe it is a vega-lite regression. Just for future reference, I found [an .. interesting .. workaround](https://github.com/altair-viz/altair/issues/1892) for `timeUnit="hoursminutes"`, which works, but it doesn't work for `timeUnit="hours"`.

PR #56, opened 2021-04-11
Overwrite beliefs

Flix6x:


timely_beliefs/beliefs/classes.py
```diff
@@ -226,15 +226,23 @@ def add_to_session(
...
-        :param session: the database session to use
-        :param beliefs_data_frame: the BeliefsDataFrame to be persisted
-        :param commit_transaction: if True, the session is committed
-                                   if False, you can still add other data to the session
-                                   and commit it all within an atomic transaction
+        If you are adding lots of beliefs, it's most efficient to use expunge_session=True and allow_overwrite=False
+
+        :param session:             the database session to use
+        :param beliefs_data_frame:  the BeliefsDataFrame to be persisted
+        :param expunge_session:     if True, all instances are removed from the session before adding beliefs
```
nhoening, line 239: 1. We need to explain a bit better in which situations this can be useful, maybe after a discussion.
2. Why does one need to flush, and does that need to happen before or after?
3. This also leads to using `bulk_save_objects` which has a few consequences that matter, so it's worth mentioning. And why do we make that distinction?

nhoening (changes requested):
I need more clarity.

nhoening:
Did you study https://docs.sqlalchemy.org/en/14/orm/persistence_techniques.html#bulk-operations-caveats before using `bulk_save_objects` in this context? Or can you say that you tested this in enough contexts?
TimedBeliefs have a lot of primary key fields and updates on them seem likely to not work. Also, they have relationship fields which are not supported. I'm considering making that a flag, as well.

Flix6x:
The context of usage is pretty narrow at the moment, since we explicitly expunge everything before bulk inserting the beliefs. Should we narrow it even more by forcing a commit after bulk inserting (for example, by disallowing a specific combination of our boolean options)?

Maybe you can help me out by adding some tests for the add_to_session method?

What do you mean with a flag for relationship fields?

nhoening:
With context, I mean adding beliefs and the different cases. For example, what if we change the data source of existing beliefs? Maybe then the bulk case will not work fully.

Sorry, with the flag I mean that users might get the choice if they want  to use `session.bulk_save_objects` or simply stick with `session.add`, which supports all sqlachemy .

The code is difficult to read now. I suggest this: First, if the user sets `expunge_session=True`, we expunge (no just when `allow_overwrite=False`, that's difficult). Then, the if/else block about `allow_overwrite`. In the case of `allow_overwrite=False`, we consider using bulk or add (using a new flag). If `allow_overwrite=True`, we use merge.

Flix6x:
Thanks for those comments, it looks much cleaner now. I also added tests. What I don't yet fully understand is the FlushError you get, in both new tests in some of the tested cases, when you would set `expunge_session=False` (which I currently do not test for), but other than that, I'd be happy to move on.

nhoening:
Looks very good now!

So `expunge_session=False` is not working for you, but it does fail some tests? I don't quite understand the severity of this.

Flix6x:
No tests are currently failing, because I do not check for the case `expunge_session=False`. If I would, some test cases would fail (on a not yet well understood FlushError). To conclude, we don't have the time to investigate this further right now, and `expunge_session=True` is our recommended, supported and documented work-around.

nhoening (approved):
Thanks for the tests. I believe this is very useful, and I hope the expunge_session=False case can be worked out later. I agree we don't have the capacity now.

PR #57, opened 2021-04-14
Allow setting limits on the y-axis range for event values

Flix6x:


timely_beliefs/visualization/utils.py
```diff
@@ -51,7 +53,18 @@ def plot(
...
+    if None in event_value_range:
+        event_value_range = list(event_value_range)
+        if event_value_range[0] is None:
+            # Infer minimum from data
+            event_value_range[0] = bdf["event_value"].min()
+        if event_value_range[-1] is None:
+            # Infer maximum from data
+            event_value_range[-1] = bdf["event_value"].max()
+        event_value_range = tuple(event_value_range)
+
```
nhoening, line 67: Maybe I'm overlooking something, but `event_value_range` doesn't seem to actually be applied anywhere?

Flix6x: It is passed to `graphs.value_vs_time_chart` in line 113 (previously line 100).

nhoening (approved):

PR #59, opened 2021-04-19
Fix deprecation warnings for unused optional arguments

Flix6x:
Stop giving deprecation warnings for optional arguments that are not used.

### DIFF
--- a/timely_beliefs/utils.py
+++ b/timely_beliefs/utils.py
@@ -178,7 +178,7 @@ def replace_deprecated_argument(
     """
     if required_argument is True and new_arg_val is None and deprecated_arg_val is None:
         raise ValueError(f"Missing argument: {new_arg_name}.")
-    elif deprecated_arg_val is not None:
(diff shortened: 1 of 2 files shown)


timely_beliefs/utils.py
```diff
@@ -176,11 +176,9 @@ def replace_deprecated_argument(
     """Util function for replacing a deprecated argument in favour of a new argument.
     If new_arg_val was not already set, it is set to deprecated_arg_val together with a FutureWarning.
     """
-    if required_argument and new_arg_val is None and deprecated_arg_val is None:
+    if required_argument is True and new_arg_val is None and deprecated_arg_val is None:
         raise ValueError(f"Missing argument: {new_arg_name}.")
-    elif new_arg_val is not None:
-        pass
-    else:
+    elif deprecated_arg_val is not None:
```
nhoening, line 181: Actually we could turn the `elif` condition into a more readable `if`.

nhoening (approved):

PR #61, opened 2021-04-21
Allow reading beliefs from xlsx files

Flix6x:


nhoening:
You mean you don't want to add it to the required dependencies? In that case we could check if it's installed and if not tell the user that they have to install it manually?

Flix6x:
That makes sense, but pandas does that already, raising:
```
ValueError: Your version of xlrd is 2.0.1. In xlrd >= 2.0, only the xls format is supported. Install openpyxl instead.
```
So leave it as is?

timely_beliefs/beliefs/utils.py
```diff
@@ -493,14 +493,22 @@ def read_csv(
     and the second column has to contain the event values.
     You also need to pass explicit values for the belief horizon/time and cumulative probability,
     in addition to the sensor and source.
-    Useful ddditional kwargs passed to pandas are parse_dates=True and infer_datetime_format=True.
+    Useful additional kwargs passed to pandas are parse_dates=True and infer_datetime_format=True.
```
nhoening, line 496: I believe `infer_datetime_format` is only useful in the csv case, and I'd also mention `sheet_name` as something really useful for the excel case:

```
Consult pandas documentation for which additional kwargs can be passed to pandas.read_csv or pandas.read_excel. Useful examples are parse_dates=True, infer_datetime_format=True (for read_csv) and sheet_name (sheet number or name, for read_excel).
```

nhoening (approved):
Only one suggestion for the docstring.

PR #66, opened 2021-09-09
convert knowledge times and belief times to UTC before calculating belief horizons

### LINKED ISSUE #65: add_to_session fails when timezones don't match (even in name)
In my case, `event_start` and `belief_time` had "Europe/Amsterdam" and "Europe/Brussels", respectively, then this happened:

```
"DatetimeArray subtraction must have the same timezones or no timezones". More details: {'src_module': '/home/nicolas/envs/sempras/lib/python3.8/site-packages/pandas/core/arrays/datetimes.py', 'src_linenr': 674, 'src_func': '_sub_datetime_arraylike', 'src_code': 'raise TypeError('}
```

Probably during `beliefs_data_frame.convert_index_from_belief_time_to_horizon()`.

I believe we can catch this problem with more clarity (real error message which displays the time zones) and maybe also earlier (even in `BDF.__init__`?). Not sure if it can/should be fixed.

nhoening:
fixes #65

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -847,7 +847,6 @@ class BeliefsDataFrame(pd.DataFrame):
         if "belief_horizon" in self.index.names:
             return self.index.get_level_values("belief_horizon")
         else:
-            return (self.knowledge_times. - self.belief_times).rename("belief_horizon")
             return (self.knowledge_times.tz_convert("UTC") - self.belief_times.tz_convert("UTC")).rename("belief_horizon")
     
     @property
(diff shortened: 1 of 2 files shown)


timely_beliefs/beliefs/classes.py
```diff
@@ -847,8 +847,9 @@ def belief_horizons(self) -> pd.TimedeltaIndex:
         if "belief_horizon" in self.index.names:
             return self.index.get_level_values("belief_horizon")
         else:
-            return (self.knowledge_times - self.belief_times).rename("belief_horizon")
-
+            return (self.knowledge_times. - self.belief_times).rename("belief_horizon")
```
Flix6x, line 850: Please remove this line.

Flix6x (changes requested):
Can you please also bump the version in setup.py to 1.4.4?

Flix6x (approved):

PR #68, opened 2021-09-13
Do not filter by source if source is empty list

Flix6x:
Two alternatives I can think of are raising or returning an empty BeliefsDataFrame. I'd like your opinion on this one, @nhoening.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -405,7 +405,9 @@ class TimedBeliefDBMixin(TimedBelief):
             )
 
         # Apply source filter
-        if source not in (None, []):
+        if source is []:
+            return BeliefsDataFrame(sensor=sensor, beliefs=[])
+        elif source is not None:
             sources: list = [source] if not isinstance(source, list) else source
             source_cls = sources[0].__class__
             q = q.join(source_cls).filter(cls.source_id.in_([s.id for s in sources]))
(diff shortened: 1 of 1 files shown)


nhoening:
I believe it's necessary to not care which source data is from, so I wouldn't return empty or raise.

Flix6x:
So, that case (not filtering by source) is already provided by the `source=None` option.

Let me provide an example for the empty list case. A developer might try to dynamically construct a list of sources that should be used for filtering. For example:
```
players = ["Raheem", "Zlatan", "Karim"]
source = [name for name in players if name[0] == "Z"]  # we'll select only beliefs from players whose name starts with the letter Z
```
When used in `search_session` this should only return beliefs by Zlatan, as intended.
However, for:
```
source = [name for name in players if name[0] == "X"]  # we'll select only beliefs from players whose name starts with the letter X
```
This would return beliefs from all players, instead of from none of the players.

nhoening:
So this parameter means `return_only_data_from_sources_in_this_list`. Then an empty list would mean to return nothing, and `None` means that the parameter is not used.

The former case could be informative (i.e. someone is trying to find out whether there is any such data), so raising seems not the right thing to do.

nhoening:
What if source is an empty string? I feel that also deserves a direct empty response?
Also we should document that we do this, so developers better understand this parameter.

Flix6x:
Timely beliefs doesn't force data source names to be a non-empty strings, so filtering for data sources with an empty string as a name could actually be a valid use case. Besides, the `source` argument of `search_session` is meant to be used (and is type annotated as such) with a `BeliefSource` or a list thereof, and not a string. Using it with a string should cause an exception in the line:
```
q = q.join(source_cls).filter(cls.source_id.in_([s.id for s in sources]))
```
with something along the lines of `str` not having the attribute `source_id`.

I slightly expanded the docstring (see https://github.com/SeitaBV/timely-beliefs/pull/68/commits/7fe3c190724735607b8e80b4d63c3b3adcc634f4).

nhoening (approved):

PR #69, opened 2021-09-13
Issue 60 delay loading of example beliefs data frames

Flix6x:


nhoening (approved):
Looks good to me - have you tested doing a local `python setup.py install` and then seeing if `import timely_beliefs` in a Python console still works and becomes a bit faster?

Flix6x:
`import timely_beliefs` now takes me 1.5 seconds instead of 1.9 seconds previously. A small gain.

PR #76, opened 2021-10-05
Issue 75 query most recent beliefs from database

Flix6x:


timely_beliefs/beliefs/classes.py
```diff
@@ -312,6 +322,7 @@ def search_session(  # noqa: C901  # todo: remove after removing deprecated argu
         :param beliefs_after: only return beliefs formed after this datetime (inclusive)
         :param beliefs_before: only return beliefs formed before this datetime (inclusive)
         :param source: only return beliefs formed by the given source or list of sources
+        :param most_recent_only: only return the most recent beliefs from each source about each event
```
nhoening, line 325: Let's add that this means we'll find the belief with the smallest belief horizon. To be explicit and so it's faster grasped.

timely_beliefs/beliefs/classes.py
```diff
@@ -419,6 +430,28 @@ def search_session(  # noqa: C901  # todo: remove after removing deprecated argu
...
+        # Apply most recent beliefs filter
+        if most_recent_only:
+            subq = (
+                session.query(
+                    cls.event_start,
+                    cls.source_id,
+                    func.min(cls.belief_horizon).label("most_recent_belief_horizon"),
+                )
+                .group_by(cls.event_start, cls.source_id)
+                .order_by(func.min(cls.belief_horizon).desc())
```
nhoening, line 442: I'm not sure why this `order_by` clause is needed. We should have one belief horizon per event&source combination by now.

timely_beliefs/beliefs/classes.py
```diff
@@ -419,6 +430,28 @@ def search_session(  # noqa: C901  # todo: remove after removing deprecated argu
...
+                .filter(cls.sensor_id == sensor.id)
```
nhoening, line 443: Would it be faster (and still work) if we first `filter()` and then `group_by()`? It could be that sqlachemy optimises this stuff but maybe not.

Would it work without this filter the same way? The main query has a few filters which you connect to after joining the subquery, I'm curious why this filter is actually in the subquery itself.

nhoening (approved):
I believe this is okay, there is only a few things I want to talk about, partly to understand better how you solved this.

PR #81, opened 2021-11-09
Issue 80 filter by most recent events

Flix6x:
I do need to adjust some terms I use in tests, but I prefer to do that in a separate PR together with some other ill-named tests and test variables, so this PR focuses on adding functionality only. For example, the "day-after" beliefs are indeed after the fact, but actually on the same day as knowledge time, which is too confusing in the long run. So please consider "day-after" here as just a name to set it apart from "day-ahead".

### DIFF
--- a/timely_beliefs/tests/test_belief_query.py
+++ b/timely_beliefs/tests/test_belief_query.py
@@ -305,6 +305,8 @@ def test_select_most_recent_deterministic_beliefs(
     multiple_day_ahead_beliefs_about_ex_post_time_slot_event: List[DBTimedBelief],
     multiple_day_after_beliefs_about_ex_post_time_slot_event: List[DBTimedBelief],
 ):
+    """Check db query filters for most recent beliefs, most recent events, and both at once."""
+
     # Query all beliefs for this sensor
     df = DBTimedBelief.search_session(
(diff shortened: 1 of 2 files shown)


timely_beliefs/tests/test_belief_query.py
```diff
@@ -281,15 +303,36 @@ def _test_empty_frame(time_slot_sensor):
 def test_select_most_recent_deterministic_beliefs(
```
nhoening, line 303: It seems we're doing two independent tests here, one for most recent beliefs, one for most recent events. Can we also test using both `most_recent_only` and `most_recent_events_only` in the same within-query?

nhoening (changes requested):

timely_beliefs/tests/test_belief_query.py
Flix6x: I added your requested test, plus a docstring.

Flix6x:
I forgot to mention I'll also rename the `most_recent_only` argument to `most_recent_beliefs_only`, but you might have guessed that. I'll add a deprecation warning for the former argument and will make sure it remains working, like I did previously for some other renamed arguments of `search_session`.

nhoening (approved):

PR #83, opened 2021-11-29
remove deprecated testing configs

nhoening:


Flix6x (approved):
Is there nothing replacing the information that we run our tests using pytest?

Flix6x:
Oh I see there is already something in the makefile.

nhoening:
Our developer Readme (linked from the main Readme) says how to run tests (`make test`). What it uses under the hood is not that crucial and can be read in `Makefile`.

PR #84, opened 2021-11-29
use setuptools_scm to manage the version (via git tags)

nhoening:


setup.py
```diff
@@ -45,7 +46,8 @@
         "timely_beliefs.examples",
         "timely_beliefs.visualization",
     ],
-    include_package_data=True,
+    packages=find_packages(),  # will include *.py files and some other types
```
Flix6x, line 49: I got an error here, when running `to_pypi`: `SyntaxError: keyword argument repeated`. There is already a `packages` key, so I guess you still need to remove lines 40-48?

Flix6x (changes requested):

setup.py
nhoening: Oh, that is right, good catch. We can also choose not add line 49.

nhoening: This is what `find_packages` finds:

```
['timely_beliefs', 'timely_beliefs.tests', 'timely_beliefs.sensors', 'timely_beliefs.sources', 'timely_beliefs.examples', 'timely_beliefs.visualization', 'timely_beliefs.beliefs', 'timely_beliefs.sensors.func_store']
```

nhoening: So I'd argue we keep using it, it's nice.

Flix6x (approved):

PR #87, opened 2021-12-04
Allow passing sensor ids

Flix6x:
And if you pass a sensor instance instead, the query that was there before was actually not needed.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -324,7 +324,7 @@ class TimedBeliefDBMixin(TimedBelief):
         The optional arguments represent optional filters.
         :param session: the database session to use
         :param sensor: sensor to which the beliefs pertain, or its unique sensor id
-        :param sensor_class: optionally pass the sensor (sub)class explicitly (only needed if you pass a sensor id and not a sensor); the class should be mapped to a database table
+        :param sensor_class: optionally pass the sensor (sub)class explicitly (only needed if you pass a sensor id instead of a sensor, and your sensor class is not DBSensor); the class should be mapped to a database table
         :param event_starts_after: only return beliefs about events that start after this datetime (inclusive)
         :param event_ends_before: only return beliefs about events that end before this datetime (inclusive)
         :param beliefs_after: only return beliefs formed after this datetime (inclusive)
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -405,20 +405,15 @@ class TimedBeliefDBMixin(TimedBelief):
                 raise ValueError("No such sensor")
         elif not isinstance(sensor, SensorDBMixin):
             raise ValueError(f"sensor {sensor} is not a {SensorDBMixin}")
-        event_resolution, knowledge_horizon_fnc, knowledge_horizon_par = (
-            sensor.event_resolution,
-            sensor.knowledge_horizon_fnc,
-            sensor.knowledge_horizon_par,
-        )
 
         # Get bounds on the knowledge horizon (so we can already roughly filter by belief time)
         (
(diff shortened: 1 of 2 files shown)


timely_beliefs/beliefs/classes.py
```diff
@@ -395,14 +395,20 @@ def search_session(  # noqa: C901  # todo: remove after removing deprecated argu
...
+                .filter(sensor.__class__.id == sensor)
+                .one_or_none()
             )
-            .filter(sensor.__class__.id == sensor.id)
-            .one_or_none()
+            if sensor is None:
+                raise ValueError("No such sensor")
+        elif not isinstance(sensor, SensorDBMixin):
+            raise ValueError(f"sensor {sensor} is not a {SensorDBMixin}")
+        event_resolution, knowledge_horizon_fnc, knowledge_horizon_par = (
```
nhoening, line 408: I believe we use all three of these variables only once and shortly after this line, so I find it better to just prepend "sensor." three times, instead of adding these 5 extra lines.

nhoening (changes requested):
Other than one syntactic change I ask for, I approve.

Flix6x:
I'm left with a major flaw in line 400-401, though: querying an `int` doesn't make a lot of sense. If we pass an id instead of a sensor, then we still have to pass the sensor's class explicitly.

nhoening:
And probably we should both have insisted on a test to catch this.

timely_beliefs/beliefs/classes.py
```diff
@@ -321,6 +322,7 @@ def search_session(  # noqa: C901  # todo: remove after removing deprecated argu
         The optional arguments represent optional filters.
         :param session: the database session to use
         :param sensor: sensor to which the beliefs pertain, or its unique sensor id
+        :param sensor_class: optionally pass the sensor (sub)class explicitly (it should be mapped to a database table)
```
nhoening, line 325: I wouldd add that this is only needed if you pass an ID and not a sensor.

timely_beliefs/tests/test_belief_query.py
```diff
@@ -300,6 +300,25 @@ def _test_empty_frame(time_slot_sensor):
...
+    """Check db query by sensor id, against query by sensor instance, for a non-empty dataset."""
+
+    # Query all beliefs for this sensor, using sensor instance (our reference)
+    df_by_instance = DBTimedBelief.search_session(
+        session=session, sensor=ex_post_time_slot_sensor, most_recent_only=False
+    )
+
+    # Query all beliefs for this sensor, using sensor id (our test)
+    df_by_id = DBTimedBelief.search_session(
+        session=session, sensor=ex_post_time_slot_sensor, most_recent_only=False
```
nhoening, line 316: I don't think this is passing in the ID. It's passing in the exact arguments as above.

nhoening (changes requested):

timely_beliefs/tests/test_belief_query.py
Flix6x: Aye, good catch.

timely_beliefs/tests/test_belief_query.py
```diff
@@ -313,7 +313,7 @@ def test_search_by_sensor_id(
 
     # Query all beliefs for this sensor, using sensor id (our test)
     df_by_id = DBTimedBelief.search_session(
-        session=session, sensor=ex_post_time_slot_sensor, most_recent_only=False
+        session=session, sensor=ex_post_time_slot_sensor.id, most_recent_only=False
```
nhoening, line 316: Seeing from the test, there is another condition: You only need to use this param if your sensor class is *not* `DBSensor`.

nhoening (changes requested):

nhoening (approved):

PR #91, opened 2021-12-24
Allow passing custom join targets

Flix6x:
Some warnings from FlexMeasures tests lead to this work. When applying custom filter criteria, it may be desired (or even necessary) to add additional `join` targets to the query. For example, when passing custom source filters using the subclassed `DataSource`:

> `SAWarning: SELECT statement has a cartesian product between FROM element(s) "timed_belief" and FROM element "data_source".  Apply join condition(s) between each element to resolve.`

This PR lets you pass custom join targets to the timely-beliefs query.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -326,12 +326,14 @@ class TimedBeliefDBMixin(TimedBelief):
         most_recent_only: bool = False,  # deprecated
         place_beliefs_in_sensor_timezone: bool = True,
         place_events_in_sensor_timezone: bool = True,
-        custom_filter_criteria: List[BinaryExpression] = None,
-        custom_join_targets: List[JoinTarget] = None,
+        custom_filter_criteria: Optional[List[BinaryExpression]] = None,
+        custom_join_targets: Optional[List[JoinTarget]] = None,
     ) -> "BeliefsDataFrame":
         """Search a database session for beliefs about sensor events.
 
-        The optional arguments represent optional filters.
+        The optional arguments represent optional filters, with two exceptions:
+        - sensor_class makes it possible to create a query on sensor subclasses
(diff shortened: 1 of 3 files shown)


timely_beliefs/beliefs/classes.py
```diff
@@ -318,6 +327,7 @@ def search_session(  # noqa: C901  # todo: remove after removing deprecated argu
         place_beliefs_in_sensor_timezone: bool = True,
         place_events_in_sensor_timezone: bool = True,
         custom_filter_criteria: List[BinaryExpression] = None,
+        custom_join_targets: List[JoinTarget] = None,
```
nhoening, line 330: Technically, this (and the other custom parameter above are `Optional`).

And then, the following sentence in the docstring is not correct anymore (now we have something else than filters among them):

"The optional arguments represent optional filters."

timely_beliefs/beliefs/classes.py
```diff
@@ -337,6 +347,7 @@ def search_session(  # noqa: C901  # todo: remove after removing deprecated argu
         :param place_beliefs_in_sensor_timezone: if True (the default), belief times are converted to the timezone of the sensor
         :param place_events_in_sensor_timezone: if True (the default), event starts are converted to the timezone of the sensor
         :param custom_filter_criteria: optionally pass additional filters, such as ones that rely on subclasses
+        :param custom_join_targets: optionally pass additional join targets, to accommodate filters that rely on subclasses
```
nhoening, line 350: Is this new feature maybe more powerful than that? Could we also rely on other tables than our standard set?

nhoening (approved):
I only have some comments about documentation.

Also, should we document on a higher level how to get rid of the kind of errors you encountered (and what they mean)? I guess these things can be confusing to developers who don't know the data model as we do.

timely_beliefs/beliefs/classes.py
Flix6x: Good point.

PR #93, opened 2021-12-24
Distinguish mean and median

Flix6x:
I had been using the median (middle value, cp=0.5) as a replacement for the mean (expected value), which was fine for symmetric distributions, but wrong for asymmetric distributions.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -1678,13 +1678,14 @@ class BeliefsDataFrame(pd.DataFrame):
         reference_belief_horizon: timedelta = None,
         reference_source: BeliefSource = None,
         return_expected_value: bool = False,
+        return_middle_value: bool = False,
     ) -> "BeliefsDataFrame":
         """Add a column with reference values.
-        By default the reference will be the expected value of the most recent belief held by the same source.
+        By default, the reference will be the probabilistic value of the most recent belief held by the same source.
+        To set a deterministic reference, use either return_expected_value or return_middle_value.
         Optionally, set a reference belief horizon.
         Optionally, set a reference source present in the BeliefsDataFrame.
-        These options allow to define what is considered to be true at a certain time after an event.
-        Todo: Option to set probabilistic reference values
+        These options allow defining what is considered to be true at a certain time after an event.
 
         :param reference_belief_time: optional datetime to indicate that
                the accuracy should be determined with respect to the latest belief held at this time
@@ -1692,7 +1693,8 @@ class BeliefsDataFrame(pd.DataFrame):
                the accuracy should be determined with respect to the latest belief at this duration past knowledge time
         :param reference_source: optional BeliefSource to indicate that
                the accuracy should be determined with respect to the beliefs held by the given source
-        :param return_expected_value: if True, set a deterministic reference by picking the expected value
+        :param return_expected_value: if True, set a deterministic reference by picking the mean value
+        :param return_middle_value: if True, set a deterministic reference by picking the median value
         """
 
         df = self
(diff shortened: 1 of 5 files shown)


timely_beliefs/beliefs/probabilistic_utils.py
```diff
@@ -539,8 +539,41 @@ def get_belief_at_cumulative_probability(
         return df2.head(1)
 
 
-def get_mean_belief(df: "classes.BeliefsDataFrame") -> "classes.BeliefsDataFrame":
-    """Convenience function to select the expected value."""
+def get_mean_belief(
+    df: "classes.BeliefsDataFrame", distribution: str = "uniform"
+) -> "classes.BeliefsDataFrame":
+    """Convenience function to select the expected value.
+    Assumes the data frame contains a single belief per event.
```
nhoening, line 546: Not sure if this is correct. As you are using it, `df`contains one (probabilistic) belief, and maybe that is what we should be saying here (and also assert if I'm correct).

In line 567, it seems you are using the length of `df` to decide the cumulative_probability. I might misinterpret what's going on, but it seems to me that makes sense only if `df` contains only one belief, and not many?

timely_beliefs/beliefs/classes.py
```diff
@@ -1678,21 +1678,23 @@ def set_reference_values(
...
 
         :param reference_belief_time: optional datetime to indicate that
                the accuracy should be determined with respect to the latest belief held at this time
         :param reference_belief_horizon: optional timedelta to indicate that
                the accuracy should be determined with respect to the latest belief at this duration past knowledge time
         :param reference_source: optional BeliefSource to indicate that
                the accuracy should be determined with respect to the beliefs held by the given source
-        :param return_expected_value: if True, set a deterministic reference by picking the expected value
+        :param return_expected_value: if True, set a deterministic reference by picking the mean value
+        :param return_middle_value: if True, set a deterministic reference by picking the median value
```
nhoening, line 1697: Here (and maybe in `set_reference` as well) we should document that you need to pick one of them, otherwise you run into an error at runtime.

timely_beliefs/beliefs/probabilistic_utils.py
Flix6x: One (probabilistic) belief per event is correct, I think.

- In line 551, I start to loop over the events (that is, loop over the beliefs, because there is one belief per event) to extract one list of values and one list of cumulative probabilities, for each event (/ for each belief).
- From these, in line 556, I create a distribution function for each event (/ for each belief).
- In line 563, I reduce the length of the data frame, from n_events * average_number_of_probablistic_values_per_event, to n_events * 1
- In line 564, I overwrite the cp index level (of the now smaller data frame) to cp=0.5. Technically, the mean value could be located at a different cp value, so this does need a fix. Something along the lines of:
  `dist.computeCDF(dist.getMean())`
  I'll get on it.
- Finally, in line 569, I overwrite the event_values column with the mean value for each event (/ for each belief).

timely_beliefs/beliefs/classes.py
Flix6x: `set_reference_values` can be called without picking either, in which case a probabilistic reference is returned. For example, see the `accuracy` method in `beliefs/classes.py`. Fun fact: timely beliefs is actually able to compute accuracy metrics again probabilistic references (something which I have yet to encounter even in literature).

If you actually ran into an error, please let me know what it was.

nhoening: Sorry, I meant that you cannot pick both.

nhoening (approved):

timely_beliefs/beliefs/classes.py
Flix6x: Ah, I see. Thanks. I took the opportunity to switch to a different API for these arguments.

PR #95, opened 2021-12-25
support installing 3.7 by adapting scipy version, but drop official support

nhoening:
I decided it's easy to continue supporting the installation of 3.7 for a while. But we should stop mentioning we support functionality.


Flix6x (approved):
Thanks!

PR #111, opened 2022-09-21
Let read_csv support interpreting local times for a given timezone

Flix6x:
This PR lets `read_csv` support interpreting local times for a given timezone, including inferring daylight savings time transitions.


nhoening (approved):
Looks great.
Two questions:

- Don't we need some support like this at all for the usual case of loading a BeliefDataframe? Are we ensured timezone aware data from there always?
- It seems we are not testing the loading of simple (2 or 3 column) CSV files yet. That would be nice. Extra issue? Or is it easy to load a very simple one and achieve coverage of that functionality?

Flix6x:
I added the tests, thanks for prompting me. And for the usual case of loading a `BeliefDataframe` we are ensured timezone aware data.

PR #112, opened 2022-10-10
Fix for extreme timedelta bounds

Flix6x:
Fix for using the `at_date` knowledge horizon function, used for sensors with a unique knowledge time. I also opened a [Pandas issue](https://github.com/pandas-dev/pandas/issues/49021) for comparing a `pd.Timedelta` to `timedelta.max`.

Also added a test and fixed the iso formatting of datetimes in knowledge horizon parameters.

### DIFF
--- a/timely_beliefs/tests/test_belief_query.py
+++ b/timely_beliefs/tests/test_belief_query.py
@@ -46,6 +46,7 @@ def test_query_belief_for_sensor_with_unique_knowledge_time(
         belief_before=pd.Timestamp("1990-06-01 00:00Z"),
     ).convert_index_from_belief_time_to_horizon()
     assert belief_df.belief_horizons[0] == timedelta(0)
+    assert belief_df.knowledge_horizons[0] == timedelta(days=22)
 
 
 @pytest.fixture(scope="function")
--- a/timely_beliefs/tests/test_belief_query.py
+++ b/timely_beliefs/tests/test_belief_query.py
@@ -19,24 +19,27 @@ from timely_beliefs.tests import session
 
 
 @pytest.fixture(scope="function")
-def belief_recorded_at_unique_knowledge_time(
+def beliefs_recorded_at_unique_knowledge_time(
     unique_knowledge_time_sensor: DBSensor, test_source_a: DBBeliefSource
-):
-    """Define belief about a future event at its unique knowledge time (e.g. a publication date)."""
-    belief = DBTimedBelief(
-        source=test_source_a,
-        sensor=unique_knowledge_time_sensor,
-        value=10,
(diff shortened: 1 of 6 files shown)


timely_beliefs/tests/test_belief_query.py
```diff
@@ -18,6 +18,36 @@
...
+    belief_recorded_at_unique_knowledge_time: DBTimedBelief,
+):
+    """Test query of sensor with a unique knowledge time, in combination with a belief time window."""
+    belief_df = DBTimedBelief.search_session(
+        session=session,
+        sensor=unique_knowledge_time_sensor,
+        beliefs_after=pd.Timestamp("1990-04-01 00:00Z"),
+        belief_before=pd.Timestamp("1990-06-01 00:00Z"),
+    ).convert_index_from_belief_time_to_horizon()
+    assert belief_df.belief_horizons[0] == timedelta(0)
```
nhoening, line 48: Leftfield question: Would `belief_horizons` also be equal to `timedelta(0)` if the df was empty?

timely_beliefs/tests/test_belief_query.py
```diff
@@ -18,6 +18,36 @@
...
+    unique_knowledge_time_sensor: DBSensor,
+    belief_recorded_at_unique_knowledge_time: DBTimedBelief,
+):
+    """Test query of sensor with a unique knowledge time, in combination with a belief time window."""
+    belief_df = DBTimedBelief.search_session(
+        session=session,
+        sensor=unique_knowledge_time_sensor,
+        beliefs_after=pd.Timestamp("1990-04-01 00:00Z"),
+        belief_before=pd.Timestamp("1990-06-01 00:00Z"),
+    ).convert_index_from_belief_time_to_horizon()
```
nhoening, line 47: Test for df containing one belief?

timely_beliefs/tests/test_belief_query.py
nhoening, line 48: I'm confused what you're testing here. Why would the belief horizon of the event not be 21 days?

Flix6x: Because the value was recorded exactly at the unique knowledge time of the sensor. The belief horizon is defined as the duration between belief time and knowledge time.

Flix6x: I guess I could makes this clearer in the test by also including an assert statement regarding the knowledge horizon (which should be 22 days, from midnight May 10th to midnight June 1st).

timely_beliefs/tests/test_belief_query.py
Flix6x: I added a second.

nhoening (approved):

PR #113, opened 2022-10-13
Read csv with filter by column

Flix6x:
This PR lets you select a subset of rows by filtering on specific values for specific columns. For example: `{4: 1995}` selects all rows where column 4 contains the value 1995.

FYI: New functionality is introduced until (new) line 601, and what follows after is just refactoring to reduce function complexity.


timely_beliefs/beliefs/utils.py
```diff
@@ -568,10 +589,16 @@ def read_csv(
         raise TypeError(
             f"Extension {ext} not recognized. Accepted file extensions are csv, xlsx and xls."
         )
+    if filter_by_column:
+        # Filter the read-in data
+        for col, val in filter_by_column.items():
+            df = df[df[col] == val]
+        # Remove the columns used to filter
+        df = df.drop(columns=filter_by_column.keys())
```
nhoening, line 597: This should only happen if they are not in `use_cols`.

timely_beliefs/beliefs/utils.py
```diff
@@ -534,6 +535,23 @@ def read_csv(
     In case the csv file contains multiple source names, you can pass a list of sources.
     Each source name will be replaced by the actual source.
 
+    :param path:                    Path (or url) to the csv file.
+    :param sensor:                  The sensor to which the data pertains.
+    :param source:                  Optionally, set a specific source for the read-in data.
+                                    If not set, the look_up_sources parameter must be used.
+    :param look_up_sources:         Optionally, pass a list of sources used to look up source names from the csv file.
+                                    If not set, the source parameter most be used.
```
nhoening, line 543: "most" -> "must" ?

nhoening (changes requested):

timely_beliefs/beliefs/utils.py
Flix6x: Nice catch. That would allow to filter not just by an extra column, but also by a column that is part of the beliefs data.

timely_beliefs/beliefs/utils.py
```diff
@@ -593,8 +597,14 @@ def read_csv(
...
         for col, val in filter_by_column.items():
             df = df[df[col] == val]
-        # Remove the columns used to filter
-        df = df.drop(columns=filter_by_column.keys())
+        # Remove the extra columns used to filter
+        df = df.drop(
+            columns=[
+                col
+                for col in filter_by_column.keys()
+                if col not in kwargs.get("usecols", [])
```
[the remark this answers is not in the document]
Flix6x: Thanks! I added another test case for filtering by an extra column that should not be in the resulting BeliefsDataFrame.

nhoening (approved):

PR #114, opened 2022-10-13
Read csv with keep_default_na

Flix6x:
I was expecting this `pd.read_csv` option to filter out `np.NaT` values, too, but saw they still slipped through.

### DIFF
--- a/timely_beliefs/tests/test_belief_io.py
+++ b/timely_beliefs/tests/test_belief_io.py
@@ -97,7 +97,7 @@ def test_load_beliefs_with_nat_values(csv_nat_file):
         sensor=tb.Sensor("Sensor Y"),
         source=tb.BeliefSource("Source A"),
         timezone="Europe/Amsterdam",
-        na_values=[""],
+        na_values=[""],  # tells pd.read_csv to treat empty cells as NaN values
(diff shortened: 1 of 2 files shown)


timely_beliefs/tests/test_belief_io.py
```diff
@@ -69,6 +84,25 @@ def test_load_beliefs(csv_file):
...
+    The relevant row should be skipped, as requested using the combination of na_values and keep_default_na.
+    """
+
+    # Load beliefs with tb.read_csv
+    df = tb.read_csv(
+        "test_NaT.csv",
+        sensor=tb.Sensor("Sensor Y"),
+        source=tb.BeliefSource("Source A"),
+        timezone="Europe/Amsterdam",
+        na_values=[""],
```
nhoening, line 100: Do we need this line here? I believe there is no na values as we skip them.

nhoening (approved):

timely_beliefs/tests/test_belief_io.py
Flix6x: We do. This tells `pd.read_csv` to treat empty cells as NaN values.

PR #118, opened 2022-10-31
Resample beliefs about instantaneous sensors

Flix6x:
This PR implements resampling for a `BeliefDataFrame` describing an instantaneous sensor. It does so by updating the data frequency without changing the event resolution. At some point, I'd like to start introducing terminology to distinguish between the resolution of events and the frequency of the data, but not in this PR.


timely_beliefs/tests/test_belief_utils.py
```diff
@@ -31,3 +33,55 @@ def test_propagate_multi_sourced_deterministic_beliefs():
             == pd.Timestamp("2000-01-01 01:00:00+00:00")
         ].droplevel("belief_time"),
     )
+
+
+@pytest.mark.parametrize(
```
nhoening, line 38: Please include to begin with some regular case. I feel you only tested the special DST transition cases.

timely_beliefs/tests/test_belief_utils.py
```diff
@@ -31,3 +33,55 @@ def test_propagate_multi_sourced_deterministic_beliefs():
...
+            "2022-03-27 01:00+01",
+            7,
+            "PT2H",
+            [2, 3, 5, 7],
+        ),  # DST transition from +01 to +02 (spring forward, contracted event)
+        (
+            "2022-10-30 01:00+02",
+            7,
+            "PT2H",
+            [2, 5, 7],
```
nhoening, line 51: I there no 3 in this one due to drop_duplicates?

timely_beliefs/beliefs/classes.py
```diff
@@ -1347,6 +1347,15 @@ def resample_events(
             return self
         df = self
 
+        # Resample instantaneous sensors
+        # The event resolution stays zero, but the data frequency updates
+        if df.event_resolution == timedelta(0):
+            index_names = df.index.names
+            df = df.reset_index().set_index("event_start")
```
nhoening, line 1354: Is this index reset dance something which the util function could offer?

timely_beliefs/beliefs/utils.py
```diff
@@ -749,3 +751,21 @@ def extreme_timedeltas_not_equal(
     if isinstance(td_a, pd.Timedelta):
         td_a = td_a.to_pytimedelta()
     return td_a != td_b
+
+
+def downsample_first(df: pd.DataFrame, resolution: timedelta) -> pd.DataFrame:
```
nhoening, line 756: This function name was zero helpful to me. Comment also didn't help a lot.

Is this downsampling, and in case of non-obvious cases opting for the first event?

Also, handling DST correctly is so crucial, we could even consider it in to he name.

timely_beliefs/beliefs/utils.py
```diff
@@ -749,3 +751,21 @@ def extreme_timedeltas_not_equal(
...
+    """Resample data representing instantaneous events.
```
nhoening, line 757: This assumption (instantaneous events) is never checked. Or that the index is even a DateTimeIndex

timely_beliefs/beliefs/utils.py
```diff
@@ -749,3 +751,21 @@ def extreme_timedeltas_not_equal(
...
+
+    Updates the data frequency, while keeping the event resolution.
+
+    Note that the data frequency may not be constant due to DST transitions
+    The duration between observations is longer for the fall DST transition,
+    and shorter for the spring DST transition.
+    """
+    ds_index = df.index.floor(
+        resolution, ambiguous=[True] * len(df), nonexistent="shift_forward"
```
nhoening, line 766: Why can we assume all our datetimes are DST?

nhoening (changes requested):
This is probably super useful, but can be mych clearer.

timely_beliefs/beliefs/utils.py
Flix6x: The `True` value is only used for ambiguous times (see https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DatetimeIndex.floor.html). Imagining that times moves from left to right, this means that for the DST transition in fall we fall back and wait out the 'ambiguous' fold on the left side before continuing to the next event, which creates an extended event. For the DST transition in spring, we spring forward and hop over the 'nonexistent' gap to continue with the next event on the right side, which creates a contracted event.

Sorry for all the terminology. For your convenience:
- fold and gap: used by the [Python datetime module](https://peps.python.org/pep-0495/#terminology)
- fall and spring: names of the seasons in which these DST transitions occur
- fall back and spring forward: a popular mnemonic to remember what happens when
- ambiguous and nonexistent: used by the Pandas package to describe the effects when needing to [localize](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DatetimeIndex.tz_localize.html) timezone naive datetimes around DST transitions (also relevant to the [floor](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DatetimeIndex.floor.html) operation)
- extended event and contracted event: used by timely-beliefs to describe the effect on the resolution of events when resampling events that include a DST transition

nhoening: We do assume that all our datetimes are ambiguous, then? No matter if DST transitions are present or relevant in the data?

Flix6x: No, the ambiguous rule is only applies to datetimes that are ambiguous. But nevertheless the array length has to match the frame length.

nhoening: Here, `True` is applied to all datetimes, ambiguous or not.

Flix6x: Can you show me what you mean, because I'm printing out the downsampled results and I'm seeing what I expect: that is, mixed offsets, instead of just DST offsets (which is +02 for our tests in the Europe/Amsterdam timezone).

I know that we are passing a `True` value to all datetimes, but pandas is not applying that value to any datetime that is not ambiguous. See the docs for the floor operation, parameter `ambiguous`, third bullet point:
> - bool-ndarray where True signifies a DST time, False designates a non-DST time (note that this flag is only applicable for ambiguous times)

Therefore, I see no point in first determining which **non**-ambiguous datetimes are DST and which are non-DST.

timely_beliefs/tests/test_belief_utils.py
Flix6x: No. Actually, you prompted me to discover that the call to `drop_duplicates` is no longer needed since I started using the `join` operation in the line following it.

Events on either side of the DST transition are 2 hours apart, and fall on even hours on either side, so midnight, 2 AM, 4 AM etc. Here there is no 3 because the event starting at 2 AM (local time, +02) is 3 hours apart from the event starting at 4 AM (local time, +01).

timely_beliefs/beliefs/utils.py
Flix6x: Thanks for raising this issue. For me it boils down to a more fundamental question: does resampling affect the data frequency or the event resolution? Hopefully these examples illustrate the question:

Let's say there is a sensor recording the average wind speed over a period of 3 seconds. That is, its event resolution is 3 seconds. And let's say a record is made every second, so we get a sliding window of measurements. That is, its data frequency is 1 second. (To be explicit, I'm following pandas terminology here, which measures frequency in units of time, say, `s` rather than `1/s`.)

Now we want to resample and save the results to a new sensor that records the maximum 3-second wind gust in a 10 minute period, for every sequential period of 10 minutes (this is actually a useful measure, e.g. reported by [KNMI](https://www.knmi.nl/research/weather-climate-models/updates/estimating-wind-gusts-during-severe-wind-conditions)). That is, the new sensor has an event resolution of 10 minutes, and also a data frequency of 10 minutes.

This resampling operation is downsampling, with both the event resolution and the data frequency being downsampled to 10 minutes. Something like `.resample(pd.Timedelta("PT10M")).max()` should do the trick to transform the data, and the event resolution would be updated to 10 minutes.

Now consider a sensor recording instantaneous temperature readings, once per hour. That is, its event resolution is 0 hours and its data frequency is 1 hour. What would it mean to resample to 30 minutes? I see two possibilities:
1. We measure instantaneous temperature every 30 minutes. The event resolution stays 0 hours and the data frequency becomes 30 minutes. Something like `.resample(pd.Timedelta("PT30M")).ffill()` (or `.interpolate()`) would do the trick to transform the data. The event resolution would not change. We are upsampling (with regards to the data frequency).
2. We measure average temperatures over a period of 30 minutes. Some combination of `.resample().interpolate().rolling().mean()` should do the trick to transform the data. The event resolution would be updated to 30 minutes. We are upsampling (with regards to the data frequency) and downsampling (with regards to the event resolution).

For example, given temperature readings:
```
10 °C at 1 PM
12 °C at 2 PM
```
With linear interpolation, option 1 yields:
```
10 °C at 1 PM
11 °C at 1.30 PM
12 °C at 2 PM
```
With linear interpolation, option 2 yields:
```
10.5 °C average between 1 PM and 1.30 PM
11.5 °C average between 1.30 PM and 2 PM
```

For now, I'm gravitating towards a policy like "in timely-beliefs, event resampling, by default, updates both the data frequency and the event resolution for non-instantaneous sensors, and only the data frequency for instantaneous sensors." But I'd like to support updating the event resolution for instantaneous sensors, too, maybe only when explicitly requested. For example, consider instantaneous battery SoC measurements every 5 minutes. When resampled to a day, one might not really be interested to know the SoC at midnight every day (`.first()`, and keep the original zero event resolution), but rather something like the daily min or max SoC (`.min()`/`.max()`, and update the event resolution to 1 day).

Flix6x:
FYI After you get a chance to reply to our open comments, I plan to clear up the terminology in the docstrings with regards to the sensor resolution vs data frequency.

timely_beliefs/beliefs/utils.py
nhoening: I understand the distinction now (between treating data frequency and event resolution)

W.r.t. your last paragraph,
1) I can imagine resampling policies, which can be chosen by way of general configuration setting or even per sensor attribute. You named two policies here, hopefully we could keep that simple for now. And you know which default you'd prefer.
2) The text you've written can probably enter the timely-beliefs documentation, so everybody can learn about our thoughts here.

Finally, you made no suggestion for a better funxtion name yet.

timely_beliefs/beliefs/utils.py
nhoening: That bracketed comment in the third bullet point in another library is easy to overlook. Can you add "pandas is not applying that value to any datetime that is not ambiguous" as a comment here?

nhoening: I also didn't read the Pandas docs like you interpreted it here. But it seems to work.

timely_beliefs/beliefs/utils.py
nhoening: `resample_to_frequency` or `resample_frequency`?

Flix6x: I renamed the function.

> I can imagine resampling policies

I now made the util function accept different resampling methods, and restricted its use in `resample_events` to BeliefsDataFrames containing as many events as data points. Solving the general case is a lot harder, but this should already support most use cases.

> The text you've written can probably enter the timely-beliefs documentation, so everybody can learn about our thoughts here.

I opened https://github.com/SeitaBV/timely-beliefs/issues/123 for this.

timely_beliefs/beliefs/utils.py
nhoening: Can you reply on this?

Flix6x: I'm more leaning towards making this a private function. Would that take away your concern?

timely_beliefs/beliefs/classes.py
```diff
@@ -1416,6 +1434,13 @@ def resample_events(
             return self
         df = self
 
+        # Resample instantaneous sensors
```
nhoening, line 1437: I can't explain why your new function only applies to instantaneous sensors. Can you add that?

And why is that the one case in which we take so much care of DST transitions?  (this `resample_events`  function here has two other cases)

Note: The NB about `keep_only_most_recent_belief=True` applies to only one case, the reader might assume it's about all of them.

timely_beliefs/beliefs/utils.py
nhoening: Not quite. If it matters so much, why not add an `assert` statement? I'm guessing you wouldn't be sure what the outcomes would be if `df` wasn't instantaneous?

nhoening (changes requested):
I approve with the logic in place here, but I still don't grok something about the instantaneous case.

timely_beliefs/beliefs/classes.py
```diff
@@ -1416,6 +1434,13 @@ def resample_events(
...
+        # The event resolution stays zero, but the event frequency is updated
```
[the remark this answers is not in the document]
Flix6x: It does here, because the resampling method defaults to `asfreq`.

Flix6x:
> Not quite. If it matters so much, why not add an `assert` statement? I'm guessing you wouldn't be sure what the outcomes would be if `df` wasn't instantaneous?

You guessed correctly. I really intended this PR to be only about resampling instantaneous sensor data. That means I just haven't investigated yet whether all of this DST handling logic would be useful for resampling non-instantaneous sensor data, too. I'll gladly make that a follow-up ticket.

One extra complexity is that the function currently works on regular DataFrames, too. They don't have a sensor as metadata.

timely_beliefs/beliefs/utils.py
Flix6x: I opened https://github.com/SeitaBV/timely-beliefs/issues/124 as a follow-up.

nhoening (approved):
Thanks for the clarifications!!

PR #119, opened 2022-10-31
Optional forecast imports

Flix6x:
Resolve import errors when installing without optional dependencies.


timely_beliefs/tests/conftest.py
```diff
@@ -12,12 +13,16 @@
...
-    collect_ignore = ["test_ignore_36.py", "test_plotting.py", "test_ridgeline.py"]
-if sys.version_info[0] == 3 and sys.version_info[1] == 7:
-    # Ignore these tests for python==3.7
-    collect_ignore = ["test_plotting.py", "test_ridgeline.py"]
+    collect_ignore += ["test_ignore_36.py"]
+if importlib.util.find_spec("altair") is None:
+    # Ignore these tests if altair is not installed
+    collect_ignore += ["test_viz__plotting.py", "test_viz__ridgeline.py"]
+if importlib.util.find_spec("sktime") is None:
+    # Ignore these tests if sktime is not installed
```
nhoening, line 24: I think I see the reasoning behind this - when someone isn't using one of our `make test[X]` functions, they'd fail these tests if they have not installed everything.

Is it a good idea to print out that this is what is happening?

Makefile
```diff
@@ -18,11 +18,16 @@ install-tb:
 	pre-commit install
 
 test:
-	pip install -e .[viz,forecast]
+	pip install -e .
 	pip install setuptools_scm pytest
-	pytest
+	pytest --ignore test_forecast --ignore test_viz
```
nhoening, line 23: Why not include (here and in the other targets) the "__" in the ignore-markers?

nhoening (approved):
Approve, just two small comments for your consideration.

PR #121, opened 2022-11-09
Print meta data

Flix6x:
This PR addresses a personal grievance. As a developer I'm often spending a little time finding out what sensor the data I'm looking at is about. From now on this information will be printed below the BeliefsDataFrame itself. This follows the Pandas convention for Series, which already prints out additional information such as:
```
Freq: H, Name: event_value, dtype: int64
```
In this PR, we start printing our custom metadata as well:
```
sensor: <Sensor: weight>, event_resolution: 0:15:00
```
If we add more metadata in the future (such as `unit`), that would then automatically show up as well.


nhoening (approved):
Very nice!

PR #122, opened 2022-11-15
Implement default resampling policy for upsampling when reading in beliefs data from CSV

Flix6x:
Sometimes you want to read in very low-res data from a CSV file and save it to a higher-res sensor (e.g. daily data saved to a hourly sensor). Not a very typical case, I would say, but I encountered it nonetheless. Usually, I believe it would point to a misconfigured sensor (too high-res), but it may also be intentional.

### DIFF
--- a/timely_beliefs/beliefs/utils.py
+++ b/timely_beliefs/beliefs/utils.py
@@ -686,10 +686,13 @@ def interpret_special_read_cases(
         if resample:
             df = df.set_index("event_start")
             if df.index.freq is None and len(df) > 2:
+                # Try to infer the event resolution from the event frequency
                 df.index.freq = pd.infer_freq(df.index)
             if df.index.freq is not None and df.index.freq > sensor.event_resolution:
+                # Upsample by forward filling
                 df = df.resample(sensor.event_resolution).ffill()
             else:
+                # Downsample by computing the mean
                 df = df.resample(sensor.event_resolution).mean()
             df = df.reset_index()
     elif len(df.columns) == 3:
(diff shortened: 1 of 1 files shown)


timely_beliefs/beliefs/utils.py
```diff
@@ -684,12 +684,14 @@ def interpret_special_read_cases(
...
-                .resample(sensor.event_resolution)
-                .mean()
-                .reset_index()
-            )
+            df = df.set_index("event_start")
+            if df.index.freq is None and len(df) > 2:
+                df.index.freq = pd.infer_freq(df.index)
+            if df.index.freq is not None and df.index.freq > sensor.event_resolution:
+                df = df.resample(sensor.event_resolution).ffill()
+            else:
```
nhoening, line 692: This else block is triggered if `df.index.freq is None and df.index.freq > sensor.event_resolution`, which is not your intention, I believe.

Better to first check `df.index.freq is not None`, then to decide if `ffill()`or `mean()` should be used?

We did not have that check in the existing code, so why do we need it now?

nhoening (changes requested):

timely_beliefs/beliefs/utils.py
Flix6x: > This else block is triggered if df.index.freq is None and df.index.freq > sensor.event_resolution, which is not your intention, I believe.

Your statement would fail on comparing None to a timedelta, but I'll take your comment as a worry that one might accidentally downsample when they wanted to upsample, in case the data frequency could not be determined (e.g. in case of irregularly spaced data). Maybe I could add a user warning in case the data frequency cannot be determined?

I already added a few inline comments to clarify the new logic.

nhoening: Thanks for clarifying - that seems to be the situation we don't want to happen.

I beklieve my point about first checking if we have `df.index.freq`  now is valid. If it's still `None` after line 690, shouldn't we print a warning and abort?

nhoening (approved):

PR #126, opened 2022-11-30
Fix tests comparing values in list

Flix6x:
- [x] Support tests comparing `None` values in lists
- [x] Checks length of lists (and fixed 1 bug recently introduced in a test)


nhoening (approved):

Flix6x:
For future reference, Python 3.10 introduces [Optional Length-Checking To zip](https://peps.python.org/pep-0618/).

PR #128, opened 2022-12-27
Read csv from file storage

Flix6x:
Attempts to supports various cases where the read-in file is in memory, but at least supports https://github.com/FlexMeasures/flexmeasures/pull/563.

### DIFF
--- a/timely_beliefs/beliefs/utils.py
+++ b/timely_beliefs/beliefs/utils.py
@@ -598,7 +598,7 @@ def read_csv(
             for col in filter_by_column.keys()
             if col not in kwargs.get("usecols", [])
         ]
-    ext = path.split(".")[-1]
+    ext = find_out_extension(path)
     if ext.lower() == "csv":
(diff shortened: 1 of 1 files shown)


timely_beliefs/beliefs/utils.py
```diff
@@ -647,6 +647,18 @@ def read_csv(
...
+def find_out_extension(path: str):
+    if isinstance(path, str):
+        ext = path.split(".")[-1]
+    elif hasattr(path, "filename"):
+        # For example, supports werkzeug.datastructures.FileStorage objects
+        ext = path.filename.split(".")[-1]
+    else:
+        # Given this function's name, we'll let Pandas attempt to read the path as a CSV file
+        ext = "csv"
+    return ext
```
nhoening, line 659: We don't do anything with `ext` but return it, so the function can be a bit more straightforward (return right away).

nhoening (approved):
Just one small improvement

PR #129, opened 2023-01-03
Resolve pandas future warnings

Flix6x:
Pandas==1.5 asks to be explicit about the `group_keys` argument and about the timezone when slicing a timezone-aware `DatetimeIndex` with another datetime.


nhoening (commented):
Why can we be sure that we are dealing with UTC times internally?

nhoening (approved):

PR #131, opened 2023-04-03
Support Pandas 2

Flix6x:
This should get us ready for upgrading to Pandas 2.

The good people of `sktime` are still working on compatibility with Pandas 2. Their latest release (0.17.1) already resolves any problems caught by our tests, but their `pyproject.toml` still suggests `pandas<2.0.0`.

### DIFF
--- a/timely_beliefs/tests/test_belief_utils.py
+++ b/timely_beliefs/tests/test_belief_utils.py
@@ -2,6 +2,7 @@ import numpy as np
 import pandas as pd
 import pytest
 
+from timely_beliefs import BeliefsDataFrame, Sensor
 from timely_beliefs.beliefs.probabilistic_utils import get_median_belief
 from timely_beliefs.beliefs.utils import (
     propagate_beliefs,
@@ -9,6 +10,16 @@ from timely_beliefs.beliefs.utils import (
 )
 from timely_beliefs.examples import get_example_df
 from timely_beliefs.tests.utils import equal_lists
+
+
+def test_propagate_metadata_on_empty_frame():
+    df = BeliefsDataFrame(sensor=Sensor("test", unit="kW"))
+    df = df.for_each_belief(get_median_belief)
+    assert df.sensor.name == "test"
+    assert df.sensor.unit == "kW"
+    df = df.groupby(level=["event_start"], group_keys=False).apply(lambda x: x.head(1))
+    assert df.sensor.name == "test"
+    assert df.sensor.unit == "kW"
 
 
 def test_propagate_multi_sourced_deterministic_beliefs():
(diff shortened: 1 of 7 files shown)


Flix6x:
Good to know: our main test suite doesn't rely on `sktime` and in this PR automatically tested against `pandas==2.0.0`. Our forecast test suite does rely on `sktime` and tested against `pandas==1.5.3`.

Flix6x:
No tips really.

I was happy to have our tests pass. Let me know if I need to check against [breaking changes](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v2.0.0.html#backwards-incompatible-api-changes) explicitly.

nhoening:
Which tests? Only the ones here, or also FlexMeasures?

nhoening:
Maybe we can use this PR to update the Python version we support, as well? This is the highest:

```
"Programming Language :: Python :: 3.9",
```

But I guess 3.10 or even 3.11 are also supported.

Flix6x:
> Which tests? Only the ones here, or also FlexMeasures?

Just the ones here. FlexMeasures still needs some attention, but that's a separate issue (and different repo), right?

Flix6x:
> Maybe we can use this PR to update the Python version we support, as well?

We test 3.10, but not 3.11. So I think I'd just add 3.10 for now.

nhoening:
Well it would add confidence here. But if course if it requires more work there, we need to see how we can manage.

nhoening:
You will do that here in this PR right?

Flix6x:
It turned out that the DataFrame constructor still needs to call `finalize` in order to propagate the metadata, so I reverted the version check for said constructor.

It is only the Series constructor that can no longer call `finalize`, otherwise I get a `RecursionError: maximum recursion depth exceeded while calling a Python object`.

timely_beliefs/tests/test_belief_utils.py
```diff
@@ -11,6 +12,16 @@
 from timely_beliefs.tests.utils import equal_lists
 
 
+def test_propagate_metadata_on_empty_frame():
+    df = BeliefsDataFrame(sensor=Sensor("test", unit="kW"))
+    df = df.for_each_belief(get_median_belief)
```
nhoening, line 17: I don't get why we compute something per belief (and later group as well) on an empty dataframe?

nhoening (commented):
Looks good, but I have one question.

timely_beliefs/tests/test_belief_utils.py
Flix6x: To test whether calling these functions on an empty frame leads to losing the frame's metadata.

PR #132, opened 2023-06-06
Support applying transformations on the BeliefsDataFrame read from file

Flix6x:
Something that came in handy when needing to convert data to the sensor unit.

### DIFF
--- a/timely_beliefs/beliefs/utils.py
+++ b/timely_beliefs/beliefs/utils.py
@@ -540,6 +540,7 @@ def read_csv(
     resample: bool = False,
     timezone: Optional[str] = None,
     filter_by_column: dict = None,
+    transformations: list[dict] = None,
     **kwargs,
 ) -> "classes.BeliefsDataFrame":
     """Utility function to load a BeliefsDataFrame from a csv file or xls sheet (see example/temperature.csv).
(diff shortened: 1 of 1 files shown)


timely_beliefs/beliefs/utils.py
```diff
@@ -644,7 +659,17 @@ def read_csv(
     if cumulative_probability is not None:
         df["cumulative_probability"] = cumulative_probability
 
-    return classes.BeliefsDataFrame(df, sensor=sensor)
+    # Construct BeliefsDataFrame
+    bdf = classes.BeliefsDataFrame(df, sensor=sensor)
+
+    # Apply transformations
+    if transformations:
```
victorgarcia98, line 666: I think we can remove this `if` by setting the default value of `transformations` to `[]` (an empty array).

victorgarcia98 (changes requested):
Nice, looks good!

I suggest a little change to simplify the code a little bit (also simplifies the input type from `Optional[list]` to `list`)

timely_beliefs/beliefs/utils.py
Flix6x: You've stumbled upon a little Python boobytrap. See https://stackoverflow.com/questions/366422/how-can-i-avoid-issues-caused-by-pythons-early-bound-default-parameters-e-g-m

victorgarcia98 (approved):

PR #133, opened 2023-06-15
Speed ups: preloading knowledge horizon functions and prefer unique() over groupby()

Flix6x:
The speed difference isn't really noticeable in our test suite, but loading larger datasets in FlexMeasures should benefit.

(It took me a while to figure out why `inspect.getmembers()` wasn't seeing our function definitions, which involved resolving a circular import by refactoring a function.)


nhoening (approved):
Good stuff!

I found no usage of  `make_deterministic`, so I don't expect speed improvements from it.

`number_of_beliefs` is used only thrice, in resample_events (lines 1494 and 1498) and form_beliefs. Not in FlexMeasures. Could bring a little speed-up when we use that functionality.

A general idea: Is [local caching on the instance](https://realpython.com/python-property/#caching-computed-attributes) a good idea? 
We'd need to introduce an attribute that flags if the instance has changed, e.g  `_mutated` for properties to decide if they indeed need to re-compute.

PR #134, opened 2023-06-15
feature: vectorize knowledge horizon function calls, and add some todos

Flix6x:
Much needed vectorization.

- [x] Need to resolve some todos.
- [x] Want to add tests for ex_post and ex_ante knowledge horizon functions. Especially the ex_post, because it is the most common. We're only testing the most difficult ones right now.

### DIFF
--- a/timely_beliefs/tests/test_belief_query.py
+++ b/timely_beliefs/tests/test_belief_query.py
@@ -159,61 +159,40 @@ def test_query_belief_with_empty_source_list(
     assert belief_df.empty
 
 
+@pytest.mark.parametrize(
+    "beliefs_after, beliefs_before, expected_length",
+    [
+        (None, None, 1),  # Just one belief set up for this sensor
+        (None, datetime(2017, 1, 1, 10, tzinfo=utc), 0),  # No beliefs a year earlier
+        (datetime(2018, 1, 3, 10, tzinfo=utc), None, 0),  # No beliefs 2 months later
+        (datetime(2018, 1, 1, 10, tzinfo=utc), None, 1),  # One belief after 10am UTC
+        (None, datetime(2018, 1, 1, 9, tzinfo=utc), 0),  # No beliefs an hour earlier
+        (datetime(2018, 1, 1, 13, tzinfo=utc), None, 0),  # No beliefs after 1pm UTC
+    ],
+)
 def test_query_belief_by_belief_time(
     ex_ante_economics_sensor: DBSensor,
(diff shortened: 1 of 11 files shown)


timely_beliefs/tests/test_belief_query.py
```diff
@@ -56,13 +56,13 @@ def test_query_belief_for_sensor_with_unique_knowledge_time(
 
 
 @pytest.fixture(scope="function")
```
victorgarcia98, line 58: Probably, this fixture should go in conftest.py.

timely_beliefs/tests/test_belief_query.py
```diff
@@ -187,7 +187,7 @@ def test_query_belief_by_belief_time(
         len(
```
victorgarcia98, line 187: Isn't `len(bdf.index) == 0` equivalent to `bdf.empty`?

victorgarcia98: Also, I think the test could be simplified using a for loop to test for different `belief_time` or the use of `pytest.parametrize`.

timely_beliefs/tests/test_belief_query.py
```diff
@@ -430,17 +430,17 @@ def test_select_most_recent_deterministic_beliefs(
 
 
 def test_select_most_recent_probabilistic_beliefs(
-    ex_post_time_slot_sensor: DBSensor,
-    multiple_probabilistic_day_ahead_beliefs_about_ex_post_time_slot_event: list[
+    ex_ante_economics_sensor: DBSensor,
```
victorgarcia98, line 433: Not sure about this, but I think that  `economics_sensor` should be `economic_sensor`, denoting that it's a sensor of an economic nature.

victorgarcia98 (commented):
Cool!

Just some ideas to simplify the tests.

timely_beliefs/tests/test_belief_query.py
Flix6x: I kinda liked how the words *economics* and *physics* fit together. It's only a test variable name anyway.

timely_beliefs/tests/test_belief_query.py
Flix6x: The beliefs data set up here is only used in the scope of this module, so I think it's the right place. Fixtures don't *have* to be in conftests per se.

Flix6x:
Thanks for the review. Good idea to refactor that test using `pytest.mark.parametrize`. It looks much cleaner now.

victorgarcia98 (approved):

PR #135, opened 2023-06-23
Use value_counts()

Flix6x:
This PR gets rid of another occurrence of using the `for_each_belief` method (which uses a `groupby` internally). It also makes sure that the underlying computation, which is now based on the faster `value_counts` method, is reused.

I wish I had discovered `df.droplevel("cumulative_probability").index.value_counts()` before. Likewise for Pandas' `cache_readonly`.

### DIFF
--- a/timely_beliefs/beliefs/__init__.py
+++ b/timely_beliefs/beliefs/__init__.py
@@ -119,8 +119,8 @@ class BeliefsAccessor(object):
     @property
     def number_of_probabilistic_beliefs(self) -> int:
         """Return the number of beliefs in the BeliefsDataFrame that are probabilistic (more than 1 unique value)."""
-        df = self._obj.for_each_belief(fnc=pd.DataFrame.nunique, dropna=True)
-        return len(df[df > 1].max(axis=1).dropna())
+        p_depth = self._obj.droplevel("cumulative_probability").index.value_counts()
+        return len(p_depth[p_depth > 1])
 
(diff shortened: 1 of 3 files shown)


timely_beliefs/beliefs/__init__.py
```diff
@@ -119,23 +118,22 @@ def number_of_sources(self):
...
     def percentage_of_probabilistic_beliefs(self) -> float:
         """Return the percentage of beliefs in the BeliefsDataFrame that are probabilistic (more than 1 unique value)."""
-        return self.number_of_probabilistic_beliefs / self.number_of_beliefs
+        return 1 - self.percentage_of_deterministic_beliefs
 
     @property
     def percentage_of_deterministic_beliefs(self) -> float:
         """Return the percentage of beliefs in the BeliefsDataFrame that are deterministic (1 unique value)."""
-        return 1 - self.number_of_probabilistic_beliefs / self.number_of_beliefs
+        return self.number_of_deterministic_beliefs / self.number_of_beliefs
```
victorgarcia98, line 136: In case of an empty bdf I think calling `percentage_of_deterministic_beliefs` will raise a `ZeroDivisionError` because `self.number_of_beliefs = 0`. With that, what would a valid value for the `percentage_of_deterministic_beliefs` in that event? 0 or 1...?

victorgarcia98 (approved):

PR #136, opened 2023-06-27
refactor: check Pandas version when importing

Flix6x:
Check Pandas version when importing rather than when calling relevant methods.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -661,17 +661,26 @@ class BeliefsSeries(pd.Series):
 
     _metadata = METADATA
 
-    @property
-    def _constructor(self):
-        def f(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


timely_beliefs/beliefs/classes.py
```diff
@@ -661,17 +661,26 @@ class BeliefsSeries(pd.Series):
...
-            return BeliefsSeries(*args, **kwargs)
 
-        return f
+            return f
+
+    else:
+
+        @property
+        def _constructor(self):
+            def f(*args, **kwargs):
```
victorgarcia98, line 680: Not sure if it provides any major benefit (except from, maybe, readability), but I think we can use the function `functools.partial`:

```python

@property
def _constructor(self):
    return functools.partial(BeliefsSeries)
```

victorgarcia98 (approved):
Great!

There's only one optional change request.

PR #156, opened 2023-12-06
Fix/read csv with upsampling

Flix6x:
Refactoring just to reuse existing upsampling functionality (for forward filling) in `read_csv`. Because simply doing `df.resample().ffill()` does not correctly resample the last event in the data frame.

### DIFF
--- a/timely_beliefs/beliefs/utils.py
+++ b/timely_beliefs/beliefs/utils.py
@@ -809,7 +809,9 @@ def interpret_special_read_cases(
     return df
 
 
-def resample_events(df: pd.DataFrame, sensor: "classes.Sensor", keep_nan_values=keep_nan_values) -> pd.DataFrame:
+def resample_events(
+    df: pd.DataFrame, sensor: "classes.Sensor", keep_nan_values=keep_nan_values
+) -> pd.DataFrame:
     if df.empty:
         return df
     df = df.set_index("event_start")
@@ -1039,7 +1041,9 @@ def meta_repr(
 
 
 def upsample_beliefs_data_frame(
-    df: "classes.BeliefsDataFrame" | pd.DataFrame, event_resolution: timedelta, keep_nan_values: bool = False
+    df: "classes.BeliefsDataFrame" | pd.DataFrame,
+    event_resolution: timedelta,
+    keep_nan_values: bool = False,
 ) -> "classes.BeliefsDataFrame":
     """Because simply doing df.resample().ffill() does not correctly resample the last event in the data frame.
 
@@ -1048,9 +1052,9 @@ def upsample_beliefs_data_frame(
     :param keep_nan_values:     If True, place back resampled NaN values. Drops NaN values by default.
     """
     from_event_resolution = df.event_resolution
-    resample_ratio = pd.to_timedelta(
-        to_offset(from_event_resolution)
-    ) / pd.Timedelta(event_resolution)
+    resample_ratio = pd.to_timedelta(to_offset(from_event_resolution)) / pd.Timedelta(
+        event_resolution
+    )
     if keep_nan_values:
         # back up NaN values
         unique_event_value_not_in_df = df["event_value"].abs().sum() + 1
--- a/timely_beliefs/beliefs/utils.py
+++ b/timely_beliefs/beliefs/utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import math
 import warnings
 from datetime import datetime, timedelta
 from typing import List, Optional, Union
@@ -9,6 +10,7 @@ import pandas as pd
 import pytz
 from packaging import version
 from pandas.core.groupby import DataFrameGroupBy
+from pandas.tseries.frequencies import to_offset
 
 from timely_beliefs import BeliefSource, Sensor
 from timely_beliefs import utils as tb_utils
@@ -657,7 +659,8 @@ def read_csv(  # noqa C901
     )
 
     # Exclude rows with NaN or NaT values
-    if not kwargs.get("keep_default_na", True):
+    keep_nan_values = kwargs.get("keep_default_na", True)
+    if not keep_nan_values:
         df = df.dropna()
 
     if event_ends_after:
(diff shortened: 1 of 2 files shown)


timely_beliefs/beliefs/utils.py
```diff
@@ -1030,3 +1038,68 @@ def meta_repr(
...
+    from_event_resolution = df.event_resolution
+    resample_ratio = pd.to_timedelta(to_offset(from_event_resolution)) / pd.Timedelta(
+        event_resolution
+    )
+    if keep_nan_values:
+        # back up NaN values
+        unique_event_value_not_in_df = df["event_value"].abs().sum() + 1
+        df = df.fillna(unique_event_value_not_in_df)
+    new_index = initialize_index(
+        start=df.index[0],
```
victorgarcia98, line 1063: I think this would break in case of having an empty DataFrame. In that case, I would just change the resolution and return the same object perhaps, complaining with a warning.

timely_beliefs/beliefs/utils.py
```diff
@@ -1030,3 +1038,68 @@ def meta_repr(
...
+) -> "classes.BeliefsDataFrame":
+    """Because simply doing df.resample().ffill() does not correctly resample the last event in the data frame.
+
+    :param df:                  In case of a regular pd.DataFrame, make sure to set df.event_resolution before passing it to this function.
+    :param event_resolution:    Resolution to upsample to.
+    :param keep_nan_values:     If True, place back resampled NaN values. Drops NaN values by default.
+    """
+    from_event_resolution = df.event_resolution
+    resample_ratio = pd.to_timedelta(to_offset(from_event_resolution)) / pd.Timedelta(
+        event_resolution
```
victorgarcia98, line 1056: What if `event_resolution == 0` (which I'm not sure what would mean in this context)?

Also, I'm thinking on the case of `from_event_resolution` is 0. The resample ratio would be 0 regardless of the value of `event_resolution`.

timely_beliefs/beliefs/utils.py
```diff
@@ -1030,3 +1038,68 @@ def meta_repr(
...
+    # 2020-03-29 10:40:00+02:00 NaN
+    # 2020-03-29 11:00:00+02:00 NaN
+    # 2020-03-29 11:20:00+02:00 NaN
+    # 2020-03-29 11:40:00+02:00 NaN
+    # 2020-03-29 12:00:00+02:00 2000.0
+    # 2020-03-29 12:20:00+02:00 2000.0
+    # 2020-03-29 12:40:00+02:00 NaN
+    df = df.reindex(new_index).fillna(
+        method="pad",
+        limit=math.ceil(resample_ratio) - 1 if resample_ratio > 1 else None,
```
victorgarcia98, line 1098: I think this line already covers the case of upsampling an instantaneous input time series.

victorgarcia98 (commented):
Nice!

I came up with a couple of edge-cases that might be interesting to double-check.

timely_beliefs/beliefs/utils.py
Flix6x: The case of `event_resolution == timedelta(0)` is handled elsewhere, such that this method would not be called, but I'll add a `raise` statement regardless. The case of `from_event_resolution == timedelta(0)` is already handled by forward filling the instantaneous beliefs. I'm adding a test, which also lead me to a fix (for applying this function on BeliefsDataFrames rather than regular Pandas DataFrames).

timely_beliefs/beliefs/utils.py
```diff
@@ -1030,3 +1036,66 @@ def meta_repr(
...
+    # 2020-03-29 10:20:00+02:00 1000.0
+    # 2020-03-29 10:40:00+02:00 NaN
+    # 2020-03-29 11:00:00+02:00 NaN
+    # 2020-03-29 11:20:00+02:00 NaN
+    # 2020-03-29 11:40:00+02:00 NaN
+    # 2020-03-29 12:00:00+02:00 2000.0
+    # 2020-03-29 12:20:00+02:00 2000.0
+    # 2020-03-29 12:40:00+02:00 NaN
+    df = df.reindex(new_index).fillna(
+        method="pad",
```
victorgarcia98, line 1129: The method `pad` is being replaced by `ffill` in an upcoming Panda version:

https://stackoverflow.com/a/48586122

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.fillna.html

timely_beliefs/beliefs/utils.py
```diff
@@ -1030,3 +1036,66 @@ def meta_repr(
...
+    :param event_resolution:    Resolution to upsample to.
+    :param keep_nan_values:     If True, place back resampled NaN values. Drops NaN values by default.
+    """
+    from_event_resolution = df.event_resolution
+    resample_ratio = pd.to_timedelta(
+        to_offset(from_event_resolution)
+    ) / pd.Timedelta(event_resolution)
+    if keep_nan_values:
+        # back up NaN values
+        unique_event_value_not_in_df = df["event_value"].abs().sum() + 1
```
victorgarcia98, line 1082: Can you add a comment here saying that we are flagging the positions of the NaN values from the original data with a unique number? Also, an example on how this number is generated. 

Aside, what if df was a DataFrame full of NaN, wouldn't `unique_event_value_not_in_df` be NaN?  In this case it doesn't matter as we would keep the NaN.

victorgarcia98 (approved):
Thanks for addressing my previous comments :smile: 

I have two little request but this is completed.

timely_beliefs/beliefs/utils.py
Flix6x: > Aside, what if df was a DataFrame full of NaN, wouldn't `unique_event_value_not_in_df` be NaN?  In this case it doesn't matter as we would keep the NaN.

It looks like this would also behave as we want it to. A unique number not already present in the series is created, as we can see here:
```
import numpy as np
import pandas as pd

s = pd.Series([np.NaN, np.NaN])
L1norm = s.abs().sum()
print(L1norm)  # 0
```

PR #160, opened 2023-12-27
Support SQLAlchemy 2.0.x

Ahmad-Wahid:
In this pull request, I've advanced the SQLAlchemy version from `1.3.19` to `2.0.0` and adjusted the queries to adhere to the `2.0.x` format. I have used pytest to test all the new changes.

## References:
- 1.x->2.x [Migration Guide](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html)
- SQLAlchemy Blog [post](https://www.sqlalchemy.org/blog/2023/01/26/sqlalchemy-2.0.0-released/)

### DIFF
--- a/dev/requirements.txt
+++ b/dev/requirements.txt
@@ -54,7 +54,7 @@ six==1.15.0
     #   isodate
     #   jsonschema
     #   python-dateutil
-sqlalchemy==1.3.19
+sqlalchemy==2.0.0
     # via timely-beliefs (setup.py)
 toolz==0.10.0
     # via altair
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
 
     # Database support
     "psycopg2-binary",
-    "SQLAlchemy",
+    "SQLAlchemy>=2",
 
     # Conditional
     # numpy's setup requires minimal Python versions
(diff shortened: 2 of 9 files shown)


Flix6x:
Great work!

I noticed that this PR constitutes a breaking change, as timely-beliefs will effectively require `SQLAlchemy>=2`. We can make that explicit in the `pyproject.toml`.

But before we do so, I want to ask @nhoening whether we'd need to (or prefer to) support both major versions for a while? My thought is that we'd accept breaking support for `SQLAlchemy<2`, and release as `timely-beliefs==2`. Then pip's dependency resolver should pick up on any compatibility issues. And if needed, we can always backport new tb features to the 1.x branch.

nhoening:
I agree with the plan.

It is not a classic breaking change, unless someone needs to use SQLAlchemy<2 for some reason.

Nevertheless, a good time to move TB to v2

Flix6x (approved):

Ahmad-Wahid:
I would want to test this when flexmeasures support `sqlalchemy==2.0` to see how fast the data is going to be loaded.

Flix6x:
Locally I didn't notice much of a difference, but let me know if you need anything from me to do this test properly (such as a dev release).

PR #171, opened 2024-03-12
Feature/upsample to instantaneous

Flix6x:


timely_beliefs/beliefs/utils.py
```diff
@@ -1054,22 +1054,57 @@ def meta_repr(
     )
 
 
+def convert_to_instantaneous(
```
victorgarcia98, line 1057: Isn't this function converting a instantaneous events (i.e. event_resolution = 0) to an equivalent non-instantaneous events  (i.e. event_resolution > 0)? I mean, in the end, more than converting this function is creating an equivalent representation.

timely_beliefs/tests/test_df_resampling.py
```diff
@@ -70,6 +70,24 @@ def df_4323(
...
+    test_source_a: BeliefSource,
+    test_source_b: BeliefSource,
+    df_wxyz: Callable[
+        [Sensor, int, int, int, int, Optional[datetime]], BeliefsDataFrame
+    ],
+) -> BeliefsDataFrame:
+    """Convenient BeliefsDataFrame to run tests on.
+    For a single sensor, it contains 4 events, for each of which 1 belief by 1 source, described by 1
+    deterministic value.
+    Note that the event resolution of the sensor is 15 minutes.
```
victorgarcia98, line 85: I would add that the frequency of the events is hourly.

timely_beliefs/beliefs/utils.py
```diff
@@ -1054,22 +1054,57 @@ def meta_repr(
...
+):
+    """Convert non-instantaneous events to instantaneous events.
+
+    Expects event_start as the sole index, and belief_time, source, cumulative_probability and event_value as columns.
+
+    :param df:              frame to convert
+    :param boundary_policy: 'min', 'max' or 'first'
+    """
+    df2 = df.copy()
+    df2.index = df2.index + df.event_resolution
```
victorgarcia98, line 1069: Could this be `df2.index - df.event_resolution`? This is related to the comment below.

timely_beliefs/tests/test_df_resampling.py
```diff
@@ -411,3 +429,26 @@ def test_downsample_instantaneous(df_instantaneous_8111):
...
+
+
+def test_upsample_to_instantaneous(df_4111, test_source_a: BeliefSource):
+    """Test upsampling deterministic beliefs about time slot event to instantaneous events."""
+    df = df_4111
+    df = df.resample_events(timedelta(minutes=0))
+    assert df.event_resolution == timedelta(minutes=0)
+    expected_values = [0, 0, 1000, 1000, 2000, 2000, 3000, 3000]
+    expected_event_starts = [
+        pd.Timestamp("2000-01-03T09:00+00"),
```
victorgarcia98, line 441: Shouldn't we expect the `time-slot` where the event is and the previous one to be filled-in? Instead of the following one. Is it a timely-belief convention?

victorgarcia98 (commented):
Awesome!

timely_beliefs/tests/test_df_resampling.py
Flix6x: I don't follow the rationale. Can you elaborate?

timely_beliefs/beliefs/utils.py
Flix6x: It's doing the opposite, from non-instantaneous to instantaneous, but I see how it's converting to an equivalent representation.

Flix6x:
Thanks. Did we clear things up?

By the way, one of my longstanding dreams is to support sensors with dynamic resolution, where the events have to define their own resolution (which becomes another index in the BeliefsDataFrame). That's basically the way that time series can be defined in FlexMeasures' flex-model fields (for SoC constraints).

(Todo: I still want to amend the docstring to mention the frequency is 1 hour.)

victorgarcia98:
> Thanks. Did we clear things up?
> 
> By the way, one of my longstanding dreams is to support sensors with dynamic resolution, where the events have to define their own resolution (which becomes another index in the BeliefsDataFrame). That's basically the way that time series can be defined in FlexMeasures' flex-model fields (for SoC constraints).
> 
> (Todo: I still want to amend the docstring to mention the frequency is 1 hour.)

The discussion helped, thanks!

Now it's clear for me. Before I thought that the scheduler assumed that the SOC to be at the end of the periods, instead of the start.

PR #172, opened 2024-03-19
add `at_date_annually` knowledge horizon function

victorgarcia98:
Recently, we've encountered the case of beliefs that were known at a certain day and month, annually. This PR introduces a new knowledge horizon function which takes a `day` and a `month` as parameters and let's you define a date in which the event was known.

If it happens that the `event_start` is before the publication year at that year, we take the date of previous year as reference.

Regarding testing, I've covered DST and leap years.

### DIFF
--- a/timely_beliefs/sensors/func_store/knowledge_horizons.py
+++ b/timely_beliefs/sensors/func_store/knowledge_horizons.py
@@ -65,8 +65,8 @@ def x_years_ago_at_date(
     if get_bounds:
         # The minimum corresponds to an event at the 1st of January and a publication date on the 31st of December on the year `x` years ago.
         # The maximum corresponds to an event just before new year's midnight and a publication date on the 1st of January on the year `x` years ago.
-        return timedelta(days=(x - 1) * MIN_DAYS_IN_A_YEAR), timedelta(
-            days=(x + 1) * MAX_DAYS_IN_A_YEAR + 1
+        return timedelta(days=(x - 1) * MIN_DAYS_IN_A_YEAR + 1), timedelta(
+            days=(x + 1) * MAX_DAYS_IN_A_YEAR
         )
 
     if isinstance(event_start, datetime):
--- a/timely_beliefs/sensors/func_store/knowledge_horizons.py
+++ b/timely_beliefs/sensors/func_store/knowledge_horizons.py
@@ -37,21 +37,21 @@ def at_date(
 
 def x_years_ago_at_date(
     event_start: datetime | pd.DatetimeIndex,
+    x: int,
     day: int,
     month: int,
-    x: int = 1,
+    z: str,
     get_bounds: bool = False,
-    z: str | None = None,
 ) -> timedelta | pd.TimedeltaIndex | tuple[timedelta, timedelta]:
     """Compute the sensor's knowledge horizon to represent the event could be known since some date, `x` years ago.
 
     For example, it can be used for a tax rate that changes annually and with a known publication date.
 
     :param event_start:     Start of the event, used as an anchor for determining the knowledge horizon.
+    :param x:               The number of years to shift the reference date to.
     :param day:             Reference day of the month of the annual date to compare against.
     :param month:           The month of the annual date to compare against.
-    :param x:               The number of years to shift the reference date to.
-    :param z:           Timezone string.
+    :param z:               Timezone string.
     :param get_bounds:      If True, this function returns bounds on the possible return value.
                             These bounds are normally useful for creating more efficient database queries when filtering by belief time.
     """
@@ -69,7 +69,7 @@ def x_years_ago_at_date(
             days=(x + 1) * MAX_DAYS_IN_A_YEAR
         )
 
-    return event_start - datetime_x_years_ago_at_date(event_start, day, month, x, z)
+    return event_start - datetime_x_years_ago_at_date(event_start, x, day, month, z)
 
 
 def ex_post(
--- a/timely_beliefs/sensors/func_store/knowledge_horizons.py
+++ b/timely_beliefs/sensors/func_store/knowledge_horizons.py
@@ -54,8 +54,12 @@ def x_years_ago_at_date(
     :param get_bounds:      If True, this function returns bounds on the possible return value.
                             These bounds are normally useful for creating more efficient database queries when filtering by belief time.
     """
+
+    if x >= 0:
+        raise ValueError("Negative value for `x` not supported.")
+
     if get_bounds:
-        return timedelta.min, timedelta.days(366 * x)
+        return timedelta(days=(x-1)*366 - 3), timedelta(days=(x+1)*366 +1)
 
     if isinstance(event_start, datetime):
         return x_years_ago_at_date_datetime(event_start, day, month, x, z)
--- a/timely_beliefs/sensors/func_store/knowledge_horizons.py
+++ b/timely_beliefs/sensors/func_store/knowledge_horizons.py
@@ -1,6 +1,7 @@
 """Function store for computing knowledge horizons given a certain event start and resolution.
 When passed get_bounds=True, these functions return bounds on the knowledge horizon,
 i.e. a duration window in which the knowledge horizon must lie (e.g. between 0 and 2 days before the event start)."""
+
 from __future__ import annotations
 
 from datetime import datetime, timedelta
@@ -62,9 +63,11 @@ def x_years_ago_at_date(
         raise ValueError("Only positive values for `x` are supported.")
 
     if get_bounds:
-        # The minimum corresponds to an event at the 1st of January and a reference on the 31st of December on the the year `x` years ago.
-        # The maximum correponds to an event on the 31st of December and a reference on the 1st of January on the year `x` years ago.
-        return timedelta(days=(x - 1) * MIN_DAYS_IN_A_YEAR), timedelta(days=(x + 1) * MAX_DAYS_IN_A_YEAR + 1)
+        # The minimum corresponds to an event at the 1st of January and a publication date on the 31st of December on the year `x` years ago.
+        # The maximum corresponds to an event just before new year's midnight and a publication date on the 1st of January on the year `x` years ago.
+        return timedelta(days=(x - 1) * MIN_DAYS_IN_A_YEAR), timedelta(
+            days=(x + 1) * MAX_DAYS_IN_A_YEAR + 1
+        )
 
     if isinstance(event_start, datetime):
         return x_years_ago_at_date_datetime(event_start, day, month, x, z)
--- a/timely_beliefs/sensors/func_store/knowledge_horizons.py
+++ b/timely_beliefs/sensors/func_store/knowledge_horizons.py
@@ -7,9 +7,8 @@ from datetime import datetime, timedelta
 
 import pandas as pd
 
-from timely_beliefs.sensors.func_store.utils import datetime_x_days_ago_at_y_oclock
+from timely_beliefs.sensors.func_store.utils import datetime_x_days_ago_at_y_oclock, x_years_ago_at_date_datetime
 
-from pytz import timezone
 
 def at_date(
     event_start: datetime | pd.DatetimeIndex | None,
@@ -52,19 +51,9 @@ def x_years_ago_at_date(
     :param z:           Timezone string.
     :param get_bounds:      If True, this function returns bounds on the possible return value.
                             These bounds are normally useful for creating more efficient database queries when filtering by belief time.
-                            In this case, the knowledge horizon is unbounded.
     """
     if get_bounds:
-        return timedelta.min, timedelta.max
-    def x_years_ago_at_date_datetime(_event_start : datetime, day : int, month : int, x : int, z : str | None) -> timedelta:
-        if z is None:
-            z = _event_start.tzinfo
-        else:
-            z = timezone(z)
-
-        anchor = dict(year=_event_start.year - x, month=month, day=day, hour=0, minute=0, second=0, microsecond=0, tzinfo=z)
-
-        return _event_start - _event_start.replace(**anchor)
+        return timedelta.min, timedelta.days(366 * x)
         
     if isinstance(event_start, datetime):
         return x_years_ago_at_date_datetime(event_start, day, month, x, z)
--- a/timely_beliefs/sensors/func_store/knowledge_horizons.py
+++ b/timely_beliefs/sensors/func_store/knowledge_horizons.py
@@ -7,7 +7,10 @@ from datetime import datetime, timedelta
 
 import pandas as pd
 
-from timely_beliefs.sensors.func_store.utils import datetime_x_days_ago_at_y_oclock, x_years_ago_at_date_datetime
+from timely_beliefs.sensors.func_store.utils import (
+    datetime_x_days_ago_at_y_oclock,
+    x_years_ago_at_date_datetime,
+)
 
 
 def at_date(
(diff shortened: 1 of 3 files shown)


timely_beliefs/sensors/func_store/test_knowledge_horizons.py
```diff
@@ -54,6 +55,59 @@ def test_fixed_knowledge_time():
...
+    assert at_date_annually(
+        event_start=datetime(2022, 11, 19, 2, tzinfo=utc),
+        **knowledge_func_params
+    ) == timedelta(days=364, hours=2) # 365 - 1
+
+    # events that occur after the reference
+    assert at_date_annually(
+        event_start=datetime(2021, 11, 21, 3, tzinfo=utc),
+        **knowledge_func_params
+    ) == timedelta(days=1, hours=3)
```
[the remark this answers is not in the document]
victorgarcia98: Indeed. Shouldn't it be `timedelta(days=366, hours=3)`, though?

timely_beliefs/sensors/func_store/test_knowledge_horizons.py
```diff
@@ -54,6 +55,59 @@ def test_fixed_knowledge_time():
     )
 
 
+def test_at_date_annually():
+    """Check definition of knowledge horizon for events known at a fixed date annually."""
```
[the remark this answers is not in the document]
victorgarcia98: True, I've added the parameter `x` to be the number of years of offset.

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +32,45 @@ def at_date(
...
+            z = timezone(z)
+
+        anchor = dict(year=_event_start.year - x, month=month, day=day, hour=0, minute=0, second=0, microsecond=0, tzinfo=z)
+
+        return _event_start - _event_start.replace(**anchor)
+        
+    if isinstance(event_start, datetime):
+        return x_years_ago_at_date_datetime(event_start, day, month, x, z)
+    else:
+        return event_start.map(lambda _event_start: x_years_ago_at_date_datetime(_event_start,day,month, x, z))
```
Flix6x, line 72: It would help speed if we can add a vectorized implementation for the DatetimeIndex case: see https://github.com/SeitaBV/timely-beliefs/blob/38e285effb8fc8521dfe235a3ea3730bcffd2952/timely_beliefs/sensors/func_store/utils.py#L38 for an example.

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +32,45 @@ def at_date(
...
+    :param month:           The month of the annual date to compare against.
+    :param x:               The number of years to shift the reference date to.
+    :param z:           Timezone string.
+    :param get_bounds:      If True, this function returns bounds on the possible return value.
+                            These bounds are normally useful for creating more efficient database queries when filtering by belief time.
+                            In this case, the knowledge horizon is unbounded.
+    """
+    if get_bounds:
+        return timedelta.min, timedelta.max
+    def x_years_ago_at_date_datetime(_event_start : datetime, day : int, month : int, x : int, z : str | None) -> timedelta:
```
Flix6x, line 59: Let's refactor this out to `func_store.utils`.

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +32,45 @@ def at_date(
...
+        if z is None:
+            z = _event_start.tzinfo
+        else:
+            z = timezone(z)
+
+        anchor = dict(year=_event_start.year - x, month=month, day=day, hour=0, minute=0, second=0, microsecond=0, tzinfo=z)
```
Flix6x, line 65: Notes for myself:

I like the defaults you added here. This makes me dream of a generic function that allows setting all of these time components, with the relative components defaulting to 0.

Very much resembling the function signature of [relativedelta](https://dateutil.readthedocs.io/en/stable/relativedelta.html) (but then with support for timezone awareness).

Most of our knowledge horizon functions then become a simpler alias for a subset of this function.

```
def ex_ante_relativedelta(years=0, months=0, weeks=0, days=0, hours=0, year=None, month=None, day=None, weekday=None, hour=None):
    tz_naive_original_time = tz_aware_original_time.tz_convert(tz).tz_localize(None)
    tz_naive_earlier_time = (tz_naive_original_time - pd.Timedelta(years=years, months=months, weeks=weeks, days=days, hours=hours))
    if hour is not None:
        tz_naive_earlier_time = tz_naive_earlier_time.floor(
            "D"
        ) + pd.Timedelta(hours=hour)
    if minute is not None:
        tz_naive_earlier_time = tz_naive_earlier_time.floor(
            "H"
        ) + pd.Timedelta(minutes=minute)
    if second is not None:
        tz_naive_earlier_time = tz_naive_earlier_time.floor(
            "T"  # or "min"
        ) + pd.Timedelta(seconds=second)
    if microsecond is not None:
        tz_naive_earlier_time = tz_naive_earlier_time.floor(
            "s"
        ) + pd.Timedelta(microseconds=microsecond)
    tz_aware_earlier_time = tz_naive_earlier_time.tz_localize(tz).tz_convert(
        original_tz
    )

from functools import partial

x_days_at_y_o_clock = partial(ex_ante_relativedelta, days=x, hour=y)
x_years_ago_at_date = partial(ex_ante_relativedelta, years=x, month=month, day=day)
```

Likewise, `def ex_post_relativedelta(event_resolution, ...)` which should anchor the relativedelta to the event start rather than to the event end.

Flix6x (changes requested):

timely_beliefs/sensors/func_store/test_knowledge_horizons.py
```diff
@@ -54,6 +55,72 @@ def test_fixed_knowledge_time():
...
+        pd.TimedeltaIndex([timedelta(days=365, hours=1), timedelta(days=364, hours=2), timedelta(days=364, hours=2), timedelta(days=366, hours=3), timedelta(days=366, hours=4)]),
+    )
+
+    knowledge_func_params_2_years = dict(month=11, day=20, x=2)    
+
+    # Check years parameter
+    assert x_years_ago_at_date(
+        event_start=datetime(2024, 11, 19, 1, tzinfo=utc),
+        **knowledge_func_params_2_years
+    ) == timedelta(days=2*365, hours=1) # 366 days - 1
```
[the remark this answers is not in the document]
victorgarcia98: Indeed, fixed.

timely_beliefs/sensors/func_store/knowledge_horizons.py
victorgarcia98: I made a try but there is no way to replace values of a DateTimeIndex with a vectorized method. Also, I tried to use a combination of timedelta and floor, but we can't apply those with a yearly frequency.

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +32,45 @@ def at_date(
...
+    :param day:             Reference day of the month of the annual date to compare against.
+    :param month:           The month of the annual date to compare against.
+    :param x:               The number of years to shift the reference date to.
+    :param z:           Timezone string.
+    :param get_bounds:      If True, this function returns bounds on the possible return value.
+                            These bounds are normally useful for creating more efficient database queries when filtering by belief time.
+                            In this case, the knowledge horizon is unbounded.
+    """
+    if get_bounds:
+        return timedelta.min, timedelta.max
```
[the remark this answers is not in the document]
victorgarcia98: Good point! I've updated the get_bounds values.

timely_beliefs/sensors/func_store/knowledge_horizons.py
victorgarcia98: This would be nice, indeed. Sadly, `Timedelta` [doesn't support years](https://pandas.pydata.org/docs/reference/api/pandas.Timedelta.html). Nevertheless, we can find an alternative way to handle this.

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +34,39 @@ def at_date(
...
+    :param event_start:     Start of the event, used as an anchor for determining the knowledge horizon.
+    :param day:             Reference day of the month of the annual date to compare against.
+    :param month:           The month of the annual date to compare against.
+    :param x:               The number of years to shift the reference date to.
+    :param z:           Timezone string.
+    :param get_bounds:      If True, this function returns bounds on the possible return value.
+                            These bounds are normally useful for creating more efficient database queries when filtering by belief time.
+    """
+    if get_bounds:
+        return timedelta.min, timedelta.days(366 * x)
```
Flix6x, line 58: I believe the bounds should be `timedelta(days=(x-1)*366+1), timedelta(days=(x+1)*366)`, assuming x>1!

Explanation for x=1:

The smallest possible value happens with an event at midnight January 1st and a knowledge time at midnight 31 December the preceding year. -> 1 day

The largest value happens with an event just before midnight December 31st and a knowledge time at midnight January 1st the preceding year. -> nearly 2 years

Flix6x: NB: I reckon things like tax returns concerning previous years may lead to sensors with negative x, so that still needs some thinking (or otherwise explicitly raising a `NotImplementedError` for x<=0).

Flix6x (changes requested):

timely_beliefs/sensors/func_store/knowledge_horizons.py
victorgarcia98: Thanks for the explanation.
I've added both the bounds and the value check.

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +34,43 @@ def at_date(
...
+    :param event_start:     Start of the event, used as an anchor for determining the knowledge horizon.
+    :param day:             Reference day of the month of the annual date to compare against.
+    :param month:           The month of the annual date to compare against.
+    :param x:               The number of years to shift the reference date to.
+    :param z:           Timezone string.
+    :param get_bounds:      If True, this function returns bounds on the possible return value.
+                            These bounds are normally useful for creating more efficient database queries when filtering by belief time.
+    """
+
+    if x < 0:
```
Flix6x, line 58: Did you think through the case for x = 0?

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +34,43 @@ def at_date(
...
+        raise ValueError("Negative value for `x` not supported.")
+
+    if get_bounds:
+        return timedelta(days=(x - 1) * 366 - 3), timedelta(days=(x + 1) * 366 + 1)
```
Flix6x, line 62: Please add a comment for future reference, explaining the numbers used, similar to the inline comment for the bounds in `x_days_ago_at_y_oclock`.

timely_beliefs/sensors/func_store/knowledge_horizons.py
Flix6x: What about something like `.to_period('Y').to_timestamp()`?

timely_beliefs/sensors/func_store/knowledge_horizons.py
victorgarcia98: True, this one I've excluded now as it yields a negative knowledge horizon.

timely_beliefs/sensors/func_store/knowledge_horizons.py
victorgarcia98: I've added a couple of comments + made the 366 into a constant variable.

timely_beliefs/sensors/func_store/knowledge_horizons.py
victorgarcia98: Thanks! I used that in combination with `DateOffset` to vectorize it.

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +34,45 @@ def at_date(
...
+    """
+
+    MAX_DAYS_IN_A_YEAR = 366
+    MIN_DAYS_IN_A_YEAR = 364
+
+    if x <= 0:
+        raise ValueError("Only positive values for `x` are supported.")
+
+    if get_bounds:
+        # The minimum corresponds to an event at the 1st of January and a reference on the 31st of December on the the year `x` years ago.
```
Flix6x, line 65: Double *the*

Flix6x, line 65: So for x=1 this should yield 1 day, but it doesn't?

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +34,45 @@ def at_date(
...
+    :param day:             Reference day of the month of the annual date to compare against.
+    :param month:           The month of the annual date to compare against.
+    :param x:               The number of years to shift the reference date to.
+    :param z:           Timezone string.
+    :param get_bounds:      If True, this function returns bounds on the possible return value.
+                            These bounds are normally useful for creating more efficient database queries when filtering by belief time.
+    """
+
+    MAX_DAYS_IN_A_YEAR = 366
+    MIN_DAYS_IN_A_YEAR = 364
```
Flix6x, line 59: 365 as far as I know (in modern times)

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +34,45 @@ def at_date(
...
+
+    if x <= 0:
+        raise ValueError("Only positive values for `x` are supported.")
+
+    if get_bounds:
+        # The minimum corresponds to an event at the 1st of January and a reference on the 31st of December on the the year `x` years ago.
+        # The maximum correponds to an event on the 31st of December and a reference on the 1st of January on the year `x` years ago.
```
Flix6x, line 66: - corre**s**ponds
- ~on the 31st of December~ -> just before new year's midnight

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +34,45 @@ def at_date(
...
+        return timedelta(days=(x - 1) * MIN_DAYS_IN_A_YEAR), timedelta(days=(x + 1) * MAX_DAYS_IN_A_YEAR + 1)
```
Flix6x, line 67: The + 1 should move to the minimum bound.

timely_beliefs/sensors/func_store/knowledge_horizons.py
```diff
@@ -31,6 +34,45 @@ def at_date(
...
+    :param z:           Timezone string.
+    :param get_bounds:      If True, this function returns bounds on the possible return value.
+                            These bounds are normally useful for creating more efficient database queries when filtering by belief time.
+    """
+
+    MAX_DAYS_IN_A_YEAR = 366
+    MIN_DAYS_IN_A_YEAR = 364
+
+    if x <= 0:
+        raise ValueError("Only positive values for `x` are supported.")
```
Flix6x, line 62: After the following thought exercise, I believe we almost support zero and negative x out of the box.

For x=0:

- The last possible knowledge time is 31st of December, which, when combined with the first possible event in the same year, starting on January 1st, should yield a minimum knowledge horizon of -365 days (i.e. knowledge time 365 days after the event started), for a leap year.
- Likewise, the first possible knowledge time is 1st of January, which, when combined with the last possible event in the same year, starting just before new year's midnight, should yield a maximum knowledge horizon of nearly 366 days (i.e. knowledge time 366 days before the event starts), for a leap year.

For x=-1, which signifies the event could be known no sooner than some date in the year following the event start, we have:

- Event starts just before new year's midnight and knowledge time on January 1st: bound is 0 days.
- Event starts January 1st and knowledge time on December 31st the following year: bound is minus 2 full years.

Flix6x: I'll follow up later with an extension, but for now we can keep the NotImplementedError.

Flix6x (commented):
Thanks! I'll do the last couple of fixes myself tomorrow.

PR #173, opened 2024-03-21
Pop `dayfirst` attribute before read_excel.

victorgarcia98:
Pandas `read_csv` function was complaining because we were providing an unexpected keyword argument `dayfirst` and I moved the already existing `pop` call before the function call.

So far, I can't explain why it worked before. Maybe it has to do with the function signature being updated with in a more recent version which doesn't allow for extra arguments.


Flix6x (approved):
Thanks. I had introduced this bug in #170.

PR #185, opened 2024-07-15
feat: bulk save objects ignoring conflicts

victorgarcia98:
Previously, it wasn't possible to bulk save beliefs with the `allow_overwrite = True`. With the help of `ON CONFLICT DO NOTHING`, we can bulk save beliefs and ignore duplicated key conflicts.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -277,22 +277,26 @@ class TimedBeliefDBMixin(TimedBelief):
         """
         # Belief timing is stored as the belief horizon rather than as the belief time
         belief_records = (
-            beliefs_data_frame.convert_index_from_belief_time_to_horizon()
-            .reset_index()
+            beliefs_data_frame.convert_index_from_belief_time_to_horizon().reset_index()
         )
-        beliefs = [cls(sensor=beliefs_data_frame.sensor, **d) for d in belief_records.to_dict("records")]
-        
+        beliefs = [
+            cls(sensor=beliefs_data_frame.sensor, **d)
+            for d in belief_records.to_dict("records")
+        ]
+
         if expunge_session:
             session.expunge_all()
-        
+
         if bulk_save_objects:
             # serialize source and sensor
-            belief_records["source_id"] = belief_records["source"].apply(lambda x: x.id) 
+            belief_records["source_id"] = belief_records["source"].apply(lambda x: x.id)
             belief_records["sensor_id"] = belief_records.sensor.id
             belief_records = belief_records.drop(columns=["source"])
-            
+
             session.execute(
-                insert(cls).values(belief_records.to_dict("records")).on_conflict_do_nothing()
+                insert(cls)
+                .values(belief_records.to_dict("records"))
+                .on_conflict_do_nothing()
             )
         elif not allow_overwrite:
             session.add_all(beliefs)
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -299,7 +299,13 @@ class TimedBeliefDBMixin(TimedBelief):
 
             if allow_overwrite:
                 smt = smt.on_conflict_do_update(
-                    index_elements=["event_start", "belief_horizon", "source_id", "sensor_id", "cumulative_probability"],
+                    index_elements=[
+                        "event_start",
+                        "belief_horizon",
+                        "source_id",
+                        "sensor_id",
+                        "cumulative_probability",
+                    ],
                     set_=dict(event_value=smt.excluded.event_value),
                 )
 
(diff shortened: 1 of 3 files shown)


timely_beliefs/beliefs/classes.py
```diff
@@ -276,18 +277,29 @@ def add_to_session(
...
+        if bulk_save_objects:
+            # serialize source and sensor
+            belief_records["source_id"] = belief_records["source"].apply(lambda x: x.id)
+            belief_records["sensor_id"] = belief_records.sensor.id
+            belief_records = belief_records.drop(columns=["source"])
+
+            session.execute(
+                insert(cls)
+                .values(belief_records.to_dict("records"))
+                .on_conflict_do_nothing()
```
Flix6x, line 299: `if not allow_overwrite` then we should have conflicts resulting in failure, though.

timely_beliefs/beliefs/classes.py
```diff
@@ -276,18 +277,29 @@ def add_to_session(
         """
         # Belief timing is stored as the belief horizon rather than as the belief time
         belief_records = (
-            beliefs_data_frame.convert_index_from_belief_time_to_horizon()
-            .reset_index()
-            .to_dict("records")
```
Flix6x, line 281: Without this line the variable name `belief_records` no longer fits, so perhaps rename it.

Flix6x (changes requested):
Looks good! I only have two minor requests, and it seems the pipeline still fails on pre-commit.

Was this option introduced recently in SQLAlchemy? If so, perhaps we should specify a minimum requirement, too.

timely_beliefs/beliefs/classes.py
```diff
@@ -276,18 +277,29 @@ def add_to_session(
         """
         # Belief timing is stored as the belief horizon rather than as the belief time
         belief_records = (
```
nhoening, line 279: Technically not "record" yet, as you do the `to_dict("records")` part later now. Still the `beliefs_data_frame`?

timely_beliefs/beliefs/classes.py
nhoening, line 299: So this would not overwrite old objects with new ones ... would we maybe prefer the other way around?

Or even more complete, we could check `allow_overwrite` here and decide whether to do this  or [`on_conflict_do _update`](https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#sqlalchemy.dialects.postgresql.Insert.on_conflict_do_update)

It seems in that case we'd need to provide fields that need to be part of the update, but I guess we can use all?

nhoening (changes requested):
One naming comment, one idea for complete support of `allow_overwrite`

timely_beliefs/beliefs/classes.py
victorgarcia98: Thanks! I've experimented a bit with this option and I found a way to update the `event_value` on conflict. I'll push the commit in a minute :)

timely_beliefs/beliefs/classes.py
victorgarcia98: True, needs renaming

timely_beliefs/beliefs/classes.py
victorgarcia98: Thanks for the suggestion

victorgarcia98:
> Was this option introduced recently in SQLAlchemy? If so, perhaps we should specify a minimum requirement, too.

Good point. This feature was intrduced in SQLAlchemy 1.1 and the project requires SQLAlchemy >= 2.0 so we don't need to constraint it.

Flix6x (changes requested):
Great catch by @nhoening. By now I would prefer to use `on_conflict_do_nothing` over raising an exception, in case `allow_overwrite is False`.

And also I think we can now default to the faster bulk saving in the function signature. I think we used the slower option as a default because it was less prone to raise errors.

victorgarcia98:
Good idea! I've implemented the suggested changes and update the tests.

Flix6x (approved):
Thanks for addressing the failing tests.

Note that I had first resolved one of the failing tests differently. You revised `add_to_session` to make sure it does not fail if an empty frame is being saved, which was the right thing to do, while I first thought the test wasn't supposed to save an empty frame in the first place, so I had started revising the test and conftest. I reverted my changes, although there is still something odd going on there, namely this:

Due to the use of `params=[1, 2]` in the `rolling_day_ahead_beliefs_about_time_slot_events` fixture, each test that uses that fixture is being called twice as many times (in the case of `test_adding_to_session`, two sets of 4 times, due to the use of `@pytest.mark.parametrize`), where in the first round of calls, the fixture set up data from source A and none from B, and the second round the fixture set up data from source B and none from A. The test then selects only data from source B. All in all, the test setup and test cases could be clearer. But I'm leaving this for another day.

Issue #177, opened 2024-05-07
KeyError: "DataFrame should contain column named 'source'."


Flix6x:
As observed e.g. in https://github.com/SeitaBV/timely-beliefs/actions/runs/8975492372/job/24650214309.

PR #188, opened 2024-09-19
Chore/upgrade dependencies

Flix6x:
Improve dependency management, and unpin some requirements for optional dependencies.


nhoening (commented):
For [the 3.11 test](https://github.com/SeitaBV/timely-beliefs/actions/runs/10943807513/job/30384187351?pr=188), I see many errors saying " KeyError: "DataFrame should contain column named 'source'.""

nhoening (commented):
There are no dev/3.10/requirements.txt, dev/3.8/requirements.txt or dev/3.7/requirements.txt

Maybe I am overlooking something but why do the tests for these Python versions not fail on that?

Flix6x:
> There are no dev/3.10/requirements.txt, dev/3.8/requirements.txt or dev/3.7/requirements.txt
> 
> Maybe I am overlooking something but why do the tests for these Python versions not fail on that?

The pipeline doesn't install the dev requirements, but decides what to install based on the `pyproject.toml`.

Flix6x:
> Mh, #177 is already closed? Why would this PR close it again? But I see its error message pop up in the 3.11 test ....

I had closed it by addressing the issue in #178, which pinned Pandas. I thought I could unpin it in this PR, but alas.

It turns out that for some reason my local Python 3.11 tests run different dependency versions than the GitHub pipeline. I'm still not sure why. The pipeline just runs `make test`, which I'm also doing locally.

FYI, I had asked you to review just before I realized that the pipeline wasn't checking 3.11 yet. After I added 3.11 to the pipeline, I realized that Pandas still needs a pin. Tests now pass, but unfortunately, pinning Pandas also means `numpy` is not going beyond `1.26.4` for a timely-beliefs installation that uses the optional `[forecast]` dependencies (with `pip install timely-beliefs[forecast].

nhoening:
Pandas needs a on only for 3.11?

Flix6x:
> Pandas needs a on only for 3.11?

What?

nhoening:
Sorry. Is the pin only needed for 3.11?

Flix6x:
> Sorry. Is the pin only needed for 3.11?

Ah I see, because only that version failed the tests. Well, I could test for 3.12, too, perhaps. And for <3.11 it seems we could let the pin loose indeed.

Flix6x:
Sorry, it doesn't look like I can let go the pin on `pandas<2.2.2`.

nhoening:
Well, at least that is very recent.
numpy sits at a version from 7 months ago (just under version 2), why is that? I don't see a pin for it? maybe that comes via pandas?

nhoening:
Oh yeah, pandas 2.2.1 is from February, and even 2.2.2 is from April

nhoening (changes requested):
Should we not update the 3.9 requirements? It seems it uses vastly different versions (e.g. looking at pandas)

Flix6x:
Pandas 2.2.2 is the first one compatible with numpy 2, but it breaks metadata propagation for our subclasses.

Updating dev requirements is not part of this PR.

nhoening (approved):

PR #199, opened 2025-07-30
Fix: field order in index for most recent event

nhoening:
The queries we generate use sensor_id first, and it also usually makes the biggest impact in the search space.


Flix6x (approved):
Great improvement!

PR #204, opened 2025-08-19
Support using materialized view

Muhammad-Moiz626:
- [x] Implemented logic to use materialized views.
- [x] Added boolean to check if user wants materialized views or use actual table.

### DIFF
--- a/timely_beliefs/tests/test_belief_persistence.py
+++ b/timely_beliefs/tests/test_belief_persistence.py
@@ -2,8 +2,8 @@ import pandas as pd
 import pytest
 from sqlalchemy.exc import IntegrityError
 
-from timely_beliefs import DBSensor, DBBeliefSource, DBTimedBelief
-from timely_beliefs.tests import session, Session
+from timely_beliefs import DBBeliefSource, DBSensor, DBTimedBelief
+from timely_beliefs.tests import Session, session
 
 
 @pytest.mark.parametrize("replace_source", [False, True])
--- a/timely_beliefs/tests/test_belief_persistence.py
+++ b/timely_beliefs/tests/test_belief_persistence.py
@@ -18,7 +18,11 @@ def test_adding_to_session(
     test_source_a,
     test_source_b,
     test_source_without_initial_data,
+    refresh_mview,
 ):
+    # Refresh, so a materialized view search actually returns the beliefs set up above
+    # (otherwise the use_mview=True case would trivially pass, having found nothing to check)
+    refresh_mview()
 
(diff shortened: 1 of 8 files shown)


Flix6x:
If you fix the linting (https://github.com/SeitaBV/timely-beliefs/actions/runs/17101730703/job/48499864963) we should see the test suite passing.

Flix6x (changes requested):
I'm thinking we should add a test fixture that makes the mview, and add `@pytest.mark.parametrize` decorators to a couple of tests covering `search_session` to make sure we still get the expected behaviour when using the mview. Especially tests covering `most_recent_beliefs_only`, `most_recent_events_only` and `most_recent_only`.

(Within the timely-beliefs package.)

timely_beliefs/tests/test_belief_persistence.py
```diff
@@ -45,9 +46,49 @@ def test_adding_to_session(
...
+        newer_bdf = DBTimedBelief.search_session(
+            session=new_session,
+            sensor=time_slot_sensor,
+            source=(
+                test_source_without_initial_data if replace_source else test_source_b
+            ),
+            most_recent_beliefs_only=True,
+            use_materialized_view=use_mview,
+        )
+        assert newer_bdf.event_value[0] == 1000
```
Flix6x: What is this supposed to prove? Document the expectation so it shows up in case this assert ever fails, using:
```suggestion
        # Proves something
        assert newer_bdf.event_value[0] == 1000, "some reason for expecting 1000"
```
And does the most recent belief we added just copy the event value from the previously most recent belief? If so, than is this the best property to use for verification?

Flix6x (changes requested):

timely_beliefs/tests/test_belief_persistence.py
Flix6x: Addressed in 3f1eebd. It turned out the assertion proved little: the mview was never refreshed in this test, so `use_mview=True` searches found an empty view and most of the test body was trivially skipped (`bdf` was empty). The test now refreshes the view after each write, so the mview path is actually exercised, and the final assertion documents what it proves: a fresh session sees the belief revision — matched by event start rather than row position, and checked by its event value (1000), which is distinct from every value set up by the fixtures, so it can only come from the newly added belief. Also wrapped the second session in try/finally, because a failing assertion used to leave an open transaction that blocked the next test's `DROP MATERIALIZED VIEW` (that's the hanging pytest run you saw).

Flix6x (approved):

PR #213, opened 2025-12-08
Chore/resolve SingleBlockManager DeprecationWarning

Flix6x:
Deal with `DeprecationWarning: Passing a SingleBlockManager to BeliefsSeries is deprecated and will raise in a future version. Use public APIs instead.`

If we're lucky, this also deals with another warning we started encountering: `SAWarning: Object of type <TimedBelief> not in session, add operation along 'DataSource.beliefs' will not proceed (This warning originated from the Session 'autoflush' process, which was invoked automatically in response to a user-initiated operation. Consider using ``no_autoflush`` context manager if this warning happened while initializing objects.)`


nhoening (approved):
Great. Tested successfully with getting prices.

Both kinds of warnings disappeared!

PR #236, opened 2026-07-09
fix: mview predicate pushdown and correctness hardening in `search_session`

Flix6x:
Two related fixes to how `search_session` uses the materialized view backing `most_recent_beliefs_only` queries: a predicate-pushdown performance fix, and a hardening commit closing two correctness holes where the mview could silently drop events.

## 1. Push event window filters down into the mview subquery

The mview join only filtered the materialized view by `sensor_id` (plus `event_start < cutoff` in the live-tail union branch), without repeating the caller's `event_starts_after` / `event_ends_before` / `event_starts_before` / `event_ends_after` bounds. Postgres does not propagate range predicates across the equijoin, so every windowed query forced a full scan (and hash join) of the entire mview for that sensor.

`apply_event_timing_filters` now accepts an optional `event_start` column, and is applied to the mview select in both the mview-only branch and the live-tail UNION branch (combined with the existing `event_start < cutoff` predicate there).

Measured on a FlexMeasures dev db (3.6M-row mview, 365-day window): 755 ms before (Parallel Seq Scan over 3.2M mview rows, hash join spilling 32 batches to disk) vs 311 ms after (Bitmap Index Scan, ~35K rows). In the FlexMeasures query path, this took the mview from consistently *losing* to the fallback (up to 5.8x slower at a 30-day window) to beating it (1.4–1.5x faster).

## 2. Disable the mview when filters could change which belief is "most recent"

The mview caches the *unfiltered* global `MIN(belief_horizon)` per `(sensor, event, source)`, while the fallback subquery applies filters *before* taking the minimum. Whenever a filter can exclude the globally-most-recent belief, joining against the mview silently drops the whole event from the result, instead of returning the most recent belief among the eligible rows. Two eligibility gaps allowed that to happen:

- **Belief-time filters**: `beliefs_after` / `beliefs_before` can be set while the `most_recent_beliefs_only` branch proceeds (for sensors with ex-ante/ex-post knowledge functions), and the eligibility check only looked at `horizons_at_least/at_most`. The mview is now also disabled when belief-time filters are set.
- **Row-level custom filter criteria**: `custom_filter_criteria` are applied before the `MIN` in the fallback, but cannot be applied to the mview (they reference beliefs-table columns the view does not have). A criterion is safe only if its truth value cannot vary within one `(event_start, source_id)` group — i.e. it references no beliefs-table columns other than the group-constant `sensor_id` / `event_start` / `source_id` (columns of other tables, such as the source table, are group-constant via the `source_id` join). A new `_custom_criteria_are_group_constant` helper introspects each criterion with `sqlalchemy.sql.visitors.iterate` and disables the mview otherwise.

The introspection is deliberately conservative and fail-safe: anything it cannot positively verify (e.g. `text()` clauses) disables the mview — a performance fallback, never a correctness risk — with a debug log line for observability. Callers with source-level criteria (like FlexMeasures' source-type filters) keep the optimization automatically, and no caller-declared trust flag is needed.

Also removes a stale "This includes any custom filters" sentence from `apply_event_timing_filters`' docstring (custom criteria are applied in `apply_belief_timing_filters`).

## Tests

- `test_select_most_recent_beliefs_with_event_window` — windowed `most_recent_beliefs_only` query, result equality between mview and non-mview paths.
- `test_most_recent_beliefs_with_belief_time_filter_bypasses_mview` — with `beliefs_before` set on an ex-post sensor, both paths agree.
- `test_most_recent_beliefs_with_row_level_custom_criterion_bypasses_mview` — constructed so the globally-most-recent belief fails the criterion, i.e. it genuinely exercises the missing-events bug.
- `test_custom_criteria_are_group_constant_helper` — unit coverage of the introspection helper (source-table columns, group-constant beliefs columns, row-level columns, `text()`, mixed `and_()`).

Full run of `test_belief_query.py` + `test_belief_persistence.py`: 83 passed.

https://claude.ai/code/session_01S4jYQMv87WJt1Y1SB7gi1w

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -474,32 +474,41 @@ class TimedBeliefDBMixin(TimedBelief):
             get_bounds=True,
         )
 
-        def apply_event_timing_filters(q):
+        def apply_event_timing_filters(q, event_start_col=None):
             """Apply filters that concern the event time.
 
-            This includes any custom filters
+            This includes any custom filters.
+
+            By default, filters are applied to the beliefs table's event_start column.
+            Pass event_start_col to apply the same (event_start-only) bounds to another
+            selectable that has an event_start column of its own, such as the
+            materialized view backing most_recent_beliefs_only queries. This lets those
+            bounds be pushed down into the materialized view subquery, so Postgres can
+            use its indexes instead of scanning (and hash-joining) the entire view.
             """
+            if event_start_col is None:
+                event_start_col = cls.event_start
             if not pd.isnull(event_starts_after):
-                q = q.filter(cls.event_start >= event_starts_after)
+                q = q.filter(event_start_col >= event_starts_after)
             if not pd.isnull(event_ends_after):
                 if sensor.event_resolution == timedelta(0):
                     # inclusive
-                    q = q.filter(cls.event_start >= event_ends_after)
+                    q = q.filter(event_start_col >= event_ends_after)
                 else:
                     # exclusive
                     q = q.filter(
-                        cls.event_start > event_ends_after - sensor.event_resolution
+                        event_start_col > event_ends_after - sensor.event_resolution
                     )
             if not pd.isnull(event_starts_before):
                 if sensor.event_resolution == timedelta(0):
                     # inclusive
-                    q = q.filter(cls.event_start <= event_starts_before)
+                    q = q.filter(event_start_col <= event_starts_before)
                 else:
                     # exclusive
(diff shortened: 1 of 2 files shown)


nhoening:
Are materialized views already on main in timely beliefs?

Flix6x:
> Are materialized views already on main in timely beliefs?

Yes, and already released in v4.0.0. They are not used unless the mview is actually created by the platform using timely-beliefs, or the mview is explicitly passed to `search_session`.

PR #242, opened 2026-07-31
Let subclasses choose the type of the value columns

Flix6x:
## What

Adds a `value_column_type` hook on `TimedBeliefDBMixin` so a host application can choose the SQLAlchemy type of the two per-row numeric value columns, `cumulative_probability` and `event_value`.

The default is unchanged (`Float`, i.e. `double precision`/float8 on PostgreSQL), so existing hosts see no difference in behaviour or DDL.

## Why

These two are the only per-row numeric columns on the beliefs table, which for most hosts is by far the largest table. Storing them as `real` (float4, 4 bytes, ~7 significant decimal digits) instead of `double precision` (float8, 8 bytes) halves their width, shrinking the table on disk by roughly 15-20%.

Whether that trade is acceptable is a deployment-specific call — ~7 significant digits is ample for instantaneous power readings but not for cumulative meter readings in the millions — so this is opt-in rather than a change of default.

Downstream, FlexMeasures wants exactly this (FlexMeasures/flexmeasures#2331, FlexMeasures/flexmeasures#2332).

## Why a hook, and not "just override the columns in your subclass"

That was the original approach in the FlexMeasures PR, and it has a non-obvious problem: **mixin columns are ordered by creation order**, so redeclaring these two columns in a subclass body moves them to the front of the table *and* of the primary key.

Measured on the actual mixin:

| | primary key column order |
|---|---|
| today (plain `Column`s) | `source_id, event_start, belief_horizon, cumulative_probability, sensor_id` |
| subclass-body override | `cumulative_probability, source_id, event_start, belief_horizon, sensor_id` |
| **this hook** | `source_id, event_start, belief_horizon, cumulative_probability, sensor_id` |

The override makes the near-constant `cumulative_probability` (almost always `0.5`) the leading primary-key column — the worst available ordering — and makes a fresh `create_all()` disagree with any existing migrated table. Declaring the columns via `declared_attr` keeps column and primary-key order byte-identical to today, for the default type and a narrowed one alike.

There is a second, quieter reason to put this upstream: a subclass-body override silently discards whatever this library declares on those columns. If `timely_beliefs` later adds a `server_default`, a check constraint or an index there, the host drops it with no error and no failing test.

## Changes

- `beliefs/classes.py`: add `value_column_type` (default `Float`); declare `cumulative_probability` and `event_value` via `declared_attr` so subclasses can customize the type. Documented inline, including the `sum(real)` caveat.
- `tests/test_db_value_column_type.py`: new, DB-free — asserts the default stays float8, that the hook narrows both columns while preserving `nullable`/`primary_key`/`default=0.5`, and that column and primary-key order do not depend on the type.
- `docs/db.md`: new "Value column precision" section covering the trade-offs.

## Caveats a host should read before narrowing

Spelled out in `docs/db.md`:

- Narrowing is **lossy and not reversible** — existing values are rounded on migration; widening again restores the type, not the digits.
- **Magnitudes matter more than units.** Cumulative meter readings or currency totals in the millions have far less headroom than instantaneous power values.
- `ALTER COLUMN ... TYPE` **rewrites the whole table** and rebuilds the primary key (since `cumulative_probability` is part of it) — plan a maintenance window.
- PostgreSQL's **`sum(real)` accumulates in `real`**, so aggregates over many rows can drift; cast to `double precision` inside aggregate queries. (`avg(real)` already accumulates in double precision; `min`/`max` are unaffected.)

`BeliefsDataFrame` holds these values as float64 regardless, so no in-memory dtypes change.

## Testing

Full suite against PostgreSQL 15: **228 passed**. The one failure in `test_viz__plotting.py::test_chart_creation` is pre-existing on `main` in my environment (an Altair version deprecation) and unrelated — verified by running it on a clean checkout.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -227,7 +227,36 @@ class TimedBeliefDBMixin(TimedBelief):
     """
     Mixin class for a table with beliefs.
     The fields source and sensor do not point to another table - overwrite them to make that happen.
+
+    The SQLAlchemy type of the two per-row numeric value columns (``cumulative_probability``
+    and ``event_value``) can be customized by setting ``value_column_type`` on your subclass.
     """
+
+    #: SQLAlchemy type used to declare the two per-row numeric value columns,
+    #: ``cumulative_probability`` and ``event_value``.
+    #:
+    #: Defaults to :class:`sqlalchemy.Float` (``double precision``/float8 on PostgreSQL,
+    #: 8 bytes per value). Since these are the only per-row numeric columns on what is
+    #: typically by far the largest table, hosts that do not need the full ~15-17
+    #: significant decimal digits of double precision can trade precision for storage by
+    #: setting this to ``Float(precision=24)``, which maps to PostgreSQL ``real``
+    #: (float4, 4 bytes, ~7 significant decimal digits)::
+    #:
+    #:     class MyBelief(Base, TimedBeliefDBMixin):
+    #:         __tablename__ = "my_timed_belief"
+    #:         value_column_type = Float(precision=24)
+    #:
+    #: This only affects how the columns are *declared*. Migrating an existing table is
+    #: up to the host, and narrowing the type is lossy and not reversible: stored values
+    #: are rounded to the nearest representable value. Before narrowing, check that the
+    #: reduced precision is acceptable for the full range of magnitudes you store -- a
+    #: sensor holding cumulative meter readings or currency totals in the millions has
+    #: far less headroom than one holding instantaneous power values.
+    #:
+    #: Note also that PostgreSQL's ``sum()`` over a ``real`` column accumulates in
+    #: ``real``, so aggregate queries over many rows may want to cast to
+    #: ``double precision`` first.
+    value_column_type = Float
 
     @declared_attr
     def __table_args__(cls):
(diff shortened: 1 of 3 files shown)


Flix6x:
Closing as not currently needed — the branch stays, so this can be picked up as-is if anyone does need it.

This hook was built to let FlexMeasures narrow its belief value columns to float4. That change (FlexMeasures/flexmeasures#2332) has since been closed: FlexMeasures is knowingly used to record cumulative meter readings, and float4 stops resolving small increments once the running total is large, silently and irreversibly. So the hook has no consumer today.

Worth recording what it is, in case that changes:

- `value_column_type` on `TimedBeliefDBMixin`, defaulting to `Float` so nothing changes for existing hosts, letting a subclass store `cumulative_probability` and `event_value` as `real` instead of `double precision`.
- It declares those columns via `declared_attr`, which — measured — keeps column and primary-key ordering byte-identical for the default type and a narrowed one alike. Redeclaring the columns in a subclass body instead reorders both, which is a trap worth remembering independently of float4.
- Tests and the trade-offs (lossy, not reversible; magnitudes matter more than units; `ALTER COLUMN` rewrites the table; `sum(real)` accumulates in `real`) are documented on the branch.

The generally useful part — that a subclass-body column override silently reshapes the primary key — was worth learning here and is now handled downstream in FlexMeasures/flexmeasures#2378 by pinning the key explicitly.

PR #244, opened 2026-08-03
Stop declaring redundant single-column indexes

Flix6x:
## What

Stops declaring `index=True` on `event_start` and `sensor_id` in `TimedBeliefDBMixin`. Both single-column indexes are fully covered by composite indexes the same mixin already declares:

| single-column index | covered by |
|---|---|
| `(event_start)` | `search_session_idx (event_start, sensor_id, source_id) INCLUDE (belief_horizon)` |
| `(sensor_id)` | `search_session_singleevent_idx (sensor_id, event_start)` |

A btree serves a leading-column lookup just as well as a dedicated single-column index, so neither adds anything a query can use. They only cost storage on what is usually by far the largest table in a deployment, and slow down every insert, update and delete that has to maintain them.

## Evidence

This is not theoretical. In a deployment where the beliefs table's indexes were profiled, **neither of these two had ever served a single index scan**, while every other index on the table had scan counts ranging from the hundreds of thousands into the millions — so this was not a case of statistics that had recently been reset. Together they accounted for a meaningful fraction of the table's total index footprint, for no benefit at all.

## Is dropping the `sensor_id` index safe for the cascade?

Yes. `sensor_id` carries `ForeignKey("sensor.id", ondelete="CASCADE")`, so deleting a sensor has to find the beliefs referencing it. PostgreSQL only needs to locate rows *by* `sensor_id`, and `search_session_singleevent_idx` leads with exactly that column, so the cascade stays index-driven.

The second test added here pins that property, so a future reordering of those composite indexes can't silently remove the coverage this change depends on.

## Impact on existing deployments

This changes only what the mixin **declares**. Existing databases keep their indexes until the host drops them; newly created ones simply never get them. Hosts that want to reclaim the space should drop them in their own migration, ideally concurrently:

```sql
DROP INDEX CONCURRENTLY IF EXISTS <table>_event_start_idx;
DROP INDEX CONCURRENTLY IF EXISTS <table>_sensor_id_idx;
```

(Index names vary with the host's naming convention.) Before doing so it's worth confirming on your own database that they really are unused, and checking read replicas too, since they keep their own counters:

```sql
SELECT indexrelname, pg_size_pretty(pg_relation_size(indexrelid)), idx_scan
  FROM pg_stat_user_indexes WHERE relname = '<your beliefs table>'
 ORDER BY pg_relation_size(indexrelid) DESC;

SELECT stats_reset FROM pg_stat_database WHERE datname = current_database();
```

## Testing

New DB-free tests in `test_db_indexes.py`: one asserting no single-column index is declared, one asserting every column we stopped indexing still leads a composite index. The first **fails on `main`** and passes here, which is what makes it worth having.

Full suite against PostgreSQL 15: 227 passed. (The one failure in `test_viz__plotting.py::test_chart_creation` is pre-existing on `main` in my environment — an Altair deprecation — and unrelated; deselected.)

## Context

Downstream counterpart: FlexMeasures/flexmeasures#2377 tracks consolidating this table's indexes, of which this is step 2. Independent of, but complementary to, #242.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -248,7 +248,9 @@ class TimedBeliefDBMixin(TimedBelief):
             ),
         )
 
-    event_start = Column(DateTime(timezone=True), primary_key=True, index=True)
+    # No index=True here: a single-column index on event_start would be redundant with
+    # the search_session_idx declared above, which leads with event_start.
+    event_start = Column(DateTime(timezone=True), primary_key=True)
     belief_horizon = Column(Interval(), nullable=False, primary_key=True)
     cumulative_probability = Column(
         Float, nullable=False, primary_key=True, default=0.5
@@ -257,11 +259,14 @@ class TimedBeliefDBMixin(TimedBelief):
 
     @declared_attr
     def sensor_id(cls):
+        # No index=True here either: a single-column index on sensor_id would be
+        # redundant with search_session_singleevent_idx, which leads with sensor_id.
+        # That composite also covers the ON DELETE CASCADE below, since PostgreSQL
+        # only needs to find rows by sensor_id and a leading-column match suffices.
         return Column(
             Integer(),
             ForeignKey("sensor.id", ondelete="CASCADE"),
             primary_key=True,
-            index=True,
         )
 
     @declared_attr
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -248,8 +248,6 @@ class TimedBeliefDBMixin(TimedBelief):
             ),
         )
 
-    # No index=True here: a single-column index on event_start would be redundant with
-    # the search_session_idx declared above, which leads with event_start.
     event_start = Column(DateTime(timezone=True), primary_key=True)
     belief_horizon = Column(Interval(), nullable=False, primary_key=True)
     cumulative_probability = Column(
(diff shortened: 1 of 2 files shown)


timely_beliefs/beliefs/classes.py
```diff
@@ -248,7 +248,9 @@ def __table_args__(cls):
             ),
         )
 
-    event_start = Column(DateTime(timezone=True), primary_key=True, index=True)
+    # No index=True here: a single-column index on event_start would be redundant with
+    # the search_session_idx declared above, which leads with event_start.
```
Flix6x: These comments belong in the commit message imo, not here.

timely_beliefs/beliefs/classes.py
```diff
@@ -257,11 +259,14 @@ def __table_args__(cls):
 
     @declared_attr
     def sensor_id(cls):
+        # No index=True here either: a single-column index on sensor_id would be
+        # redundant with search_session_singleevent_idx, which leads with sensor_id.
+        # That composite also covers the ON DELETE CASCADE below, since PostgreSQL
+        # only needs to find rows by sensor_id and a leading-column match suffices.
```
Flix6x: Likewise

timely_beliefs/beliefs/classes.py
Flix6x: Agreed, removed in 73c1309 — the rationale is in the commit message, and `test_composite_indexes_cover_the_columns_we_stopped_indexing` enforces the property the comment was describing, which is a better guard than prose anyway. `classes.py` is now a two-line diff.

timely_beliefs/beliefs/classes.py
Flix6x: Removed too, same commit.

PR #245, opened 2026-08-19
fix: enforce strict total ordering on BeliefSource

### LINKED ISSUE #238: BeliefSource.__lt__ is not a total order for same-name sources, letting pandas silently turn sources into NaN
`BeliefSource.__lt__` compares by string representation only:

```python
def __lt__(self, other):
    """Set a rule for ordering."""
    return self.__str__() < other.__str__()
```

For two *distinct* source objects that share a name (e.g. two versions of the same forecaster, or a user source and an unrelated script source that happen to be named alike), this is not a strict total order consistent with equality (`__eq__` is object identity):

- `a < b` is `False` and `b < a` is `False`, but also `a != b`;
- with the `@total_ordering` decorator this derives `a > b == True` **and** `b > a == True`.

## Why it matters: pandas can silently corrupt the source index level

When pandas needs to sort an object index level and then locate values in it again (e.g. when `pd.concat` unions the `source` level of several BeliefsDataFrames), it relies on `__lt__` being a strict total order. With same-name distinct sources, the sort order among them is arbitrary and lookups can miss, mapping rows to index code `-1` — which materializes as **`NaN` in the `source` index level**, silently replacing a valid source.

Deterministic reproduction (pandas 2.2.1, timely-beliefs current main):

```python
import pandas as pd
import timely_beliefs as tb
from datetime import timedelta

sensor = tb.Sensor("x", event_resolution=timedelta(hours=1))
sources = [tb.BeliefSource("s" + str(i % 2 + 1)) for i in range(6)]  # 6 objects, 2 names
E = pd.date_range("2025-01-01", periods=5, freq="1h", tz="UTC")
B = pd.date_range("2024-12-31", periods=3, freq="1h", tz="UTC")
spec = [(4, 1, 0, [0.3, 0.7]), (2, 3, 2, [0.5]), (1, 3, 2, [0.5]),
        (3, 0, 1, [0.3, 0.7]), (1, 3, 1, [0.5]), (5, 2, 0, [0.5]),
        (4, 2, 2, [0.3, 0.7])]
frames = [
    tb.BeliefsDataFrame([
        tb.TimedBelief(sensor=sensor, source=sources[s], event_start=E[e],
                       belief_time=B[b], cumulative_probability=cp, event_value=1.0)
        for cp in cps
    ])
    for s, e, b, cps in spec
]
bdf = pd.concat(frames)
print(sum(1 for t in bdf.index if not isinstance(t[2], tb.BeliefSource)), "NaN sources")
# 4 NaN sources
```

Real-world exposure: any concat/aggregation over frames whose source levels hold distinct objects with equal names, e.g. deserialized sources, or DB sources representing multiple versions of the same model (FlexMeasures hit this while testing its latest-version filtering).

## Fix

Make `__lt__` a strict total order by tie-breaking on object identity, which is consistent with the identity-based `__eq__`/`__hash__`:

```python
def __lt__(self, other):
    return (str(self), id(self)) < (str(other), id(other))
```

Verified that this makes the reproduction above yield 0 NaN sources.

taran-dev4u:
Closes #238. Enforces strict total ordering on BeliefSource by implementing __eq__, __lt__, and __hash__ methods. This prevents pandas index alignment from failing (and resulting in NaNs) when identical sources are present.

### DIFF
--- a/timely_beliefs/sources/classes.py
+++ b/timely_beliefs/sources/classes.py
@@ -32,8 +32,10 @@ class BeliefSource(object):
         return self.name
 
     def __lt__(self, other):
-        """Set a rule for ordering."""
-        return self.__str__() < other.__str__()
+        """Set a strict total order, tiebreaking on object identity when names are equal."""
+        if not isinstance(other, BeliefSource):
+            return NotImplemented
+        return (str(self), id(self)) < (str(other), id(other))
 
 
 class BeliefSourceDBMixin(BeliefSource):
--- a/timely_beliefs/tests/conftest.py
+++ b/timely_beliefs/tests/conftest.py
@@ -52,8 +52,9 @@ def db():
     session.close()
 
     # Defensive cleanup for optional mviews
-    with engine.begin() as conn:
-        conn.execute(text(DROP_MVIEW_SQL))
+    if engine.name != "sqlite":
+        with engine.begin() as conn:
+            conn.execute(text(DROP_MVIEW_SQL))
 
     Base.metadata.drop_all(engine)
 
@@ -63,6 +64,9 @@ def most_recent_beliefs_mview(db, request):
     """Create a materialized view if the use_mview fixture is used."""
     if "use_mview" not in request.fixturenames:
         return
+
+    if engine.name == "sqlite":
+        pytest.skip("SQLite does not support materialized views")
 
     conn = session.connection()
 
--- a/timely_beliefs/tests/test_source_ordering.py
+++ b/timely_beliefs/tests/test_source_ordering.py
@@ -1,0 +1,129 @@
+"""Tests for BeliefSource strict total ordering (issue #238).
+
+BeliefSource equality and hashing use object identity (Python default).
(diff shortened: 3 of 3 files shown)


Flix6x (changes requested):
I think this PR misses the main point of #238, which is to tiebreak on object identity.

taran-dev4u:
Thanks for the review @Flix6x!

You're right - the previous approach was wrong. It overrode __eq__/__hash__ to be name-based, making same-name objects *equal* to each other and hiding the problem rather than fixing it.

This update applies the minimal fix from the issue itself - tiebreaking on object identity in __lt__:

    return (str(self), id(self)) < (str(other), id(other))

__eq__ and __hash__ are now left untouched (Python default object-identity), consistent with the id()-based tiebreak in __lt__. The six tests in test_source_ordering.py are rewritten to reflect identity semantics, and include the exact pd.concat NaN reproduction from issue #238 (asserting 0 NaN sources after the fix).

Flix6x (approved):

PR #249, opened 2026-09-07
fix: align pandas 2.0+ constructor delegation to resolve SingleBlockManager deprecation (#232)

### LINKED ISSUE #35: Metadata lost on multiplication

### LINKED ISSUE #232: DeprecationWarning: Passing a SingleBlockManager to BeliefsSeries is deprecated
We use pandas (2.1) and see these warnings:

```
server-1  | /usr/local/lib/python3.12/dist-packages/timely_beliefs/beliefs/classes.py:776: DeprecationWarning: Passing a SingleBlockManager to BeliefsSeries is deprecated and will raise in a future version. Use public APIs instead.
server-1  |   super().__init__(*args, **kwargs)
```

taran-dev4u:
Fixes #232.
Fixes #35.

In pandas 2.0+, internal slicing and series operations were hitting `DeprecationWarning: Passing a SingleBlockManager to BeliefsSeries is deprecated` because `BeliefsDataFrame._constructor` and `_constructor_sliced` returned wrapper functions rather than delegating directly to `partial(BeliefsDataFrame)` / `partial(BeliefsSeries)`. This updates `_constructor`, `_constructor_sliced`, and `_constructor_expanddim` to use the pandas 2.0+ constructor pattern and updates `__finalize__` to avoid overwriting existing metadata with `None` during arithmetic operations.

Added unit tests covering `BeliefsSeries` metadata initialization, copying, and verified deprecation warnings are no longer emitted on slicing and series ops.

### DIFF
--- a/timely_beliefs/beliefs/classes.py
+++ b/timely_beliefs/beliefs/classes.py
@@ -5,10 +5,7 @@ import math
 import operator
 import types
 from datetime import datetime, timedelta
-from functools import partial
 from typing import TYPE_CHECKING, Any, Callable, Literal, Type, Union
-
-from packaging import version
 
 if TYPE_CHECKING:
     import altair as alt
@@ -67,6 +64,24 @@ JoinTarget = Union[
     AliasedClass,
     types.FunctionType,
 ]
+
+
+def is_pandas_manager(obj) -> bool:
+    """Check if an object is an internal pandas BlockManager or ArrayManager."""
+    return (
+        hasattr(obj, "blocks")
+        or type(obj).__name__
+        in (
+            "SingleBlockManager",
+            "SingleArrayManager",
+            "BlockManager",
+            "ArrayManager",
+        )
+        or getattr(getattr(obj, "__class__", None), "__module__", "").startswith(
+            "pandas.core.internals"
+        )
+    )
+
 
 # Columns of the beliefs table that are constant within one (event_start, source_id)
 # group, and can therefore be safely referenced by a custom filter criterion that is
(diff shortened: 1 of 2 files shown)


Flix6x:
Thanks for picking this up, and sorry it sat here without any CI. That was our fault: `pull_request` was not a trigger on our workflow, so a pull request from a fork never ran a single check. Fixed in #250, and I have merged `main` into your branch so that CI now reports here.

First, about the two issues you linked, because I think we owe you a correction there rather than the other way around. Both are ours, and both look stale. #35 is mine from 2020, and #232 is from earlier this year. I cannot reproduce either one on current `main`, on any pandas version I tried: 1.5.3, 2.0.3, 2.1.1, 2.1.4, 2.2.3 and 2.3.3. For #232 I probed `iloc`, `head`, `copy`, arithmetic, `to_frame`, `groupby.mean`, `unstack`, `resample` and DataFrame slicing, and saw no `SingleBlockManager` warning anywhere. For #35, `bdf * -1` keeps its sensor throughout. The `_constructor_from_mgr` and `_constructor_expanddim_from_mgr` methods already on `main` were added for exactly this, and they appear to have settled both.

So you were working from issues that we should have closed a while ago. That is on us, and I will verify and close them separately.

On the change itself, the suite now fails on all five Python versions:

- `test_belief_io.py::test_groupby_retains_metadata[empty_df]`
- `test_belief_utils.py::test_propagate_metadata_on_empty_frame`

I ran the same comparison locally across the six pandas versions above. `main` is green on all of them; this branch is green only on 1.5.3, and fails those same two tests on every version from 2.0 upwards.

The cause looks structural rather than incidental. For pandas >= 2.0 the change replaces the `__finalize__`-wrapping `_constructor` and `_constructor_sliced` with a bare `partial(...)`, which leaves `_constructor_from_mgr` as the only path that still carries metadata. Pandas builds empty frames through the plain constructor rather than the manager path, so `sensor` and `event_resolution` come out as `None` there. That is also why 1.5.3 is unaffected: it takes the pre-2.0 branch, which the change leaves intact. The `if val is not None` guards added to `__finalize__` cannot recover this, because they only stop good metadata being overwritten; they cannot restore what was never set, and they make it impossible to clear a field on purpose.

Two smaller notes:

- `test_beliefs_series_no_single_block_manager_deprecation` passes on unmodified `main`, so it would not catch a regression of #232.
- The `BeliefsSeries(sensor=..., event_resolution=...)` constructor keywords are a real new API, independent of both issues. If you would like to pursue that part, a separate pull request would be welcome, and we can discuss it on its own merits.

Given that neither linked issue seems to reproduce any more, I am inclined to close this one rather than ask you to keep working on it. If you are seeing the warning on a version or in a code path I have missed, please do say so, and I will reopen and take another look.

taran-dev4u:
Thanks for the detailed check @Flix6x! Makes total sense — if _constructor_from_mgr already settled both on main across modern pandas releases, closing this is definitely the right call. Appreciate you verifying it across versions!

PR #252, opened 2026-09-14
Experiment/inline independent copula

TeaDrinkingProgrammer:
Follow-up to #246. openturns was the last heavy default dependency, which was always imported, even though only 4 functions use it.

Split into `probabilistic_backend.py` (everything openturns, imported lazily, raises an `ImportError` naming the extra) and `independent_joint_cdf.py` (plain numpy, independent copula only: the joint pdf is just the product of the marginal pdfs). Custom copulas and openturns-distribution marginals still route to the openturns backend, unchanged.

**Core install (no extras):**

| | before | after |
|---|---|---|
| total site-packages | **428 MB** | **190 MB** |
| openturns (+`.libs`) | 238 MB | gone |

**Marginal cost of `import timely_beliefs`** over an interpreter that already has pandas loaded:

| | maxrss | anon |
|---|---|---|
| before | 143.9 MB | 66.2 MB |
| after | **30.3 MB** | **28.7 MB** |

Also short-circuits `probabilistic_nan_mean` when every event in the slice is a single deterministic belief. Multi-source deterministic resampling misses the fast track in `classes.py:1832` (it requires `number_of_sources == 1`) and was building openturns distributions for data that was never probabilistic: 19,200 `computeQuantile` calls for 96 deterministic values. `resample_events`, warmed, min of 5:

| case | before | after |
|---|---|---|
| 96 events, 2 sources | 415.7 ms | **132.1 ms** |
| 192 events, 2 sources | 718.7 ms | **263.0 ms** |
| 384 events, 2 sources | 1402.5 ms | **525.8 ms** |

Breaking: calling the probabilistic functions without `timely-beliefs[probabilistic]` now raises at call time, so this wants a major version bump. `interpret_complete_cdf` and `joint_cdf_with_copula` are renamed, with deprecation shims at the old names.

The numpy `independent_joint_cdf` is the part I would most like attacked. It is cross-checked against the openturns implementation in `test_probabilistic_downsampling.py`, but I don't know this maths well, and getting it wrong gives wrong statistics rather than a crash.

Note: I wrote this using an agent. I tried to check the code/SWE part as much as possible, but I can't validate the maths. Feel free to dismiss if it doesn't make sense

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,11 +34,11 @@ classifiers = [
 ]
 dependencies = [
     "importlib_metadata",
+    "packaging",
     "pytz",
     "isodate",
     # https://github.com/SeitaBV/timely-beliefs/issues/234
     "openturns>=1.23",
-    "properscoring",
 
     # Database support
     "psycopg2-binary",
@@ -54,10 +54,6 @@ dependencies = [
     "pandas >= 1.1.5, < 1.3; python_version == '3.7'",
     # https://github.com/SeitaBV/timely-beliefs/issues/148
     "pandas >= 1.4.0, != 2.1.0, !=2.1.1, < 3; python_version > '3.7'",
-    # scipy's setup requires minimal Python versions
-    "scipy<1.6; python_version <= '3.6'",
-    "scipy<1.8; python_version <= '3.7'",
-    "scipy; python_version > '3.7'",
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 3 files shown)


TeaDrinkingProgrammer:
I had a design where the openturns and the independent flow were separated by modules, but while that makes the code cleaner, that breaks the API completely.

I could make it so that we prepare for that in the code and add a deprecation warning but let me know what you think

